PR #7, opened 2011-06-27
Made digging a little more realistic. Also added BlockBreak and DamageEvents.

akrieger:
No protection against malicious clients.


SpaceManiac:
Whoops, already merged this, don't know why it's still open.

PR #9, opened 2011-07-02
Added sneaking support

zhuowei:
The sneaking support is based on lightstone's sneaking support.


SpaceManiac:
Will look over this as soon as I get the chance. I need to double-check the way the public setSneaking is intended to work as the docs are unclear about whether it is intended to freeze the player in a sneaking state or set them that way until they sneak and unsneak again.

SpaceManiac:
Double-checked - your implementation of setSneaking is correct. I'll look over this for cleanliness (it's got some complex stuff going on) when I get a chance.

zhuowei:
Basically the list that stores mob metadata is moved to GlowLivingEntity and sneaking is done by setting the flag for it in the metadata.

SpaceManiac:
Merged in f7e4463d430b2eee73f5c58421ee128073116bf5.

PR #11, opened 2011-07-05
Added /me, /time, and /kick commands

zhuowei:
These commands are built into the official server (And also built into Lightstone!), so I thought that they should be added.


SpaceManiac:
Will merge this when I do some fixes to the way Glowstone's built-in commands are handled.

PR #13, opened 2011-07-08
Added support for spawning lightning bolts

zhuowei:
Currently, they do not harm entities or set the ground on fire(on the server; the client generates client-side fire that is removed a few seconds later); they are spawned and removed 30 ticks later.
Also, when the spawn(location, class) method gets implemented, this might need to be updated to use it. (interestingly, CraftBukkit does not appear to use the spawn method to spawn lightning.)

It currently does not dispatch the LightningStrikeEvent.


SpaceManiac:
Wasn't sure how I'd been planning to do this, but this looks good. I'll merge it shortly. Thanks!

SpaceManiac:
Merged with a little bit of cleanup in 8e9a35355e06a3adb861d8bb7a03e39dd3bbc53d. Thanks!

PR #14, opened 2011-07-09
Added the mob spawning code from CraftBukkit

zhuowei:
Pig, Chicken, Cow and Sheep can be spawned; the rest are commented out until the respective classes are made.

### DIFF
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -9,6 +9,7 @@ import java.util.Random;
 import java.util.logging.Level;
 
 import org.bukkit.BlockChangeDelegate;
+import org.bukkit.Bukkit;
 import org.bukkit.Chunk;
 import org.bukkit.ChunkSnapshot;
 import org.bukkit.Effect;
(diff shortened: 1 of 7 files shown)


SpaceManiac:
My plan is to hold out on mob classes and associated spawning code until I have something meaningful to put inside them - I've been planning the mob AI infrastructure in my head and should put it on paper sometime.

zhuowei:
What about other entities, such as boats?

SpaceManiac:
Those will probably also get work done around the same time.

akrieger:
Also I'm not sure what the licensing issues copying code can do. Ideally Glowstone would be completely separate from CB, being only an implementation of Bukkit. While it probably won't matter, copying the code over does break the clean-room reverse engineering principle which Glowstone would be following by just implementing Bukkit.

rmichela:
CB is LGPL. Glowstone seems to be licensed under a public domain license. To legally incorporate CB code into Glowstone you will have to make the project GLP or LGPL, both of which are more restrictive than your current license.

zhuowei:
Good point. I'll close this pull request then.
Also, the license is the MIT license, not public domain.

PR #16, opened 2011-07-12
Fixed the packet id of PlayEffectCommand

zhuowei:
The current ID is 0x64, which is the id of the open window packet. Also, the codec now uses a fixed length buffer of length 17 instead of a dynamic buffer.


SpaceManiac:
Merged as of 9169f5fc8dbb2f1d57c9d384072c0de5e47e9ee8.

PR #15, opened 2011-07-11
Writing world data and minecraft.net name authentication

zml2008:
This adds reading of a world's level.dat and writing to a world's region file. Lighting is not saved correctly due to the lack of a proper lighting system. This also adds authentication with minecraft.net, using the same online-mode config option as the vanilla server.

PR #17, opened 2011-07-12
Changed digging to use BlockProperties, added sound effects to it

zhuowei:
Digging now looks up the drop in BlockProperties. Also, the sound effects depend on pull request #16.
Also, GlowWorld now has a playEffectExceptTo method so that we could broadcast sounds to other players without playing them to the player that made the sound.


zhuowei:
Closing this since #15 already has BlockProperties drops.

PR #18, opened 2011-07-15
Fix for block placement and breaking with damage values

zml2008:
^^


SpaceManiac:
Merged as of 9cebfd0c049876ab5892de29b3133d00b973afa5.

PR #20, opened 2011-07-16
Ebeans and config move

zml2008:
Implemented ebeans and moved config files into their own subdirectory


SpaceManiac:
Merged as of eeef4973ed8dff13615c2c428458cd74022bd0e6.

Issue #26, opened 2011-07-20
ITS OVER 9000 ERRORS!


winsock:
You told IRC to give you work to do so.
Make the version number read OVER 9000!!!!!!!!

SpaceManiac:
:P

zhuowei:
There, I fixed it for you in pull request #27.

PR #27, opened 2011-07-20
Version number now OVER 9000!

zhuowei:
This should "fix" issue #26.
@winsock, I hope you are happy.
The version number reported by the new and more powerful Glowstone: 
`This server is running Glowstone version 9001.git-Glowstone-0.0.0-133-g44093b6`

### DIFF
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -548,7 +548,7 @@ public final class GlowServer implements Server {
      * @return version of this server implementation
      */
     public String getVersion() {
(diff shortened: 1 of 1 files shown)


winsock:
I support this :D

zhuowei:
Maybe I should future-proof it by changing the version number to 9002? :)

winsock:
That might result in a StackOverflowException!!!!!

zhuowei:
Let's make sure we get the stack to overflow.

`public String getVersion(){
    return "OVER 9" + thousand();
}
private String thousand(){
    return ",000" + thousand();
}`

PR #30, opened 2011-07-26
Added sound effects for digging

zhuowei:
Also, created a new method in world called playEffectExceptTo, which plays a sound effect to everyone except one player.


SpaceManiac:
Merged as of c970d1b8cc31a3c7de37e50276c152cab0a253e3.

PR #34, opened 2011-09-03
Fixes for the changes in bukkit, a ban system + api, and a few more events called

zml2008:
PlayerPreLoginEvent, PlayerLoginEvent, and PlayerTeleportEvent were added


zml2008:
So it looks like the permissions I added for vanilla commands aren't working properly ingame (even default true). The permissions act as expected in the console though. Dunno what's causing this though.

zhuowei:
Bukkit re-implemented all of the standard commands recently ( bukkit/bukkit@0914132467dd338fa2094640deb3afcb2b7c03e5 ); maybe it's conflicting with our version?

zml2008:
This includes standard command support, so stuff should work fine.

zml2008:
Fixed the permissions problem. loadPlugins clears all the registered permissions and their defaults with pluginManager.clearPlugins(), so I changed the order of registration.

SpaceManiac:
The first five commits are now in master and the last two are now in mc-1.8. Thanks!

PR #53, opened 2011-12-09
Added stub for getAllowEnd

Raphfrk:
The current repo doesn't compile against the Bukkit that is downloaded (I wiped my maven repo).


Raphfrk:
Hmm, I should have checked for other pulls :p

PR #55, opened 2011-12-10
Using ASCII to represent UTF-8

hef:
git seems to choke on UTF-8 encoded source files.


SpaceManiac:
Changes effectively included as of c2b703ce6491faae96910deafc39b4102ce5f34a.

PR #56, opened 2011-12-10
Entity spawn code and basic chicken class

Raphfrk:
This pull adds code to allow entities to be spawned in the world.

It also adds a GlowChicken class as an example.

A map is created which maps Bukkit entity classes to GlowStone entity classes and then reflection is used to construct the entity.

There is also a ReflectHelper class that could be expanded if other reflection operations are required.


Raphfrk:
Hmm, should have checked for branches first :).  Not having much luck with pull for you guys

PR #58, opened 2011-12-11
Added dummy getKiller() methods for bukkit.

hef:


pbfy0:
Yeah, these are needed
i get a compile error. I tried to fix them, but I didn't have any luck

PR #66, opened 2014-01-09
Update build.gradle

Grinch:
- Add 'mavenLocal' to repository list.
- Specify to not include '-shadow' in the JAR name in /build/distributions/.
- Specify default tasks to perform, now you can just run 'gradle' (or 'gradlew'/'gradlew.bat')

Signed-off-by: Steven Downer grinch@outlook.com

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,6 @@
+// Tasks to perform when no task is specified
+defaultTasks 'clean', 'build', 'shadowJar'
+
 // Plugins used
 apply plugin: 'java' // Specifies a java project, allows for javadoc generation etc.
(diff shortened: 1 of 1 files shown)


SpaceManiac:
Is there a reason you duplicate JUnit as both compile and testCompile? Other than that, I think the changes look good.

lukespragg:
You don't need JUnit in the final shadow jar, so testCompile should be all you need. @Grinch?

Grinch:
Yeah, not sure why I added that. Fixed now though.

SpaceManiac:
Merged in 8b359f61e545377d32d10661ff62d610d5c10665.

PR #68, opened 2014-01-10
Remove fancy versioning

Grinch:
This prevents issues when git isn't available to gradle.

Signed-off-by: Steven Downer grinch@outlook.com


zml2008:
just do a try-catch and return "unknown" if executing git fails

PR #72, opened 2014-02-07
Switch to thread-per-world design for ticking

zml2008:
WorldScheduler manages world threads, and is triggered by the scheduler.

All Bukkit API access still goes through scheduler thread.
EventFactory syncs events to main thread (except for events marked as async)

Scheduler task execution has been completely restructured:
- No more separate GlowWorker
- Async tasks are run through an ExecutorService, not individual threads
- No more separate newTasks, oldTasks, and tasks. It's now a single ConcurrentHashMap.
- Tasks are executed _before_ world tick, not after.

More thread-safeing is left to be done, which will allow more to be
placed on alternate threads.


SpaceManiac:
Merged in 8a75ed9a91ac7a5f5eda0cd7f70dca7e7b5b334d.

PR #89, opened 2014-02-14
Made player join after having sent the chunks

xxmicloxx:
This PR makes the player join AFTER having sent the chunks... This fixes the stuck-in-floor-bug and also looks nicer.


SpaceManiac:
I made what I think is a more elegant change in c1b6db6 - it seems to work nicer for me, so let me know if it's sufficient for you.

xxmicloxx:
Yep, that's better... I just didn't know setRawLocation existed...

Issue #103, opened 2014-06-07
AsyncPlayerChatEvent


dequis:
Only PlayerChatEvent is implemented.

```
// todo: async this
PlayerChatEvent event = EventFactory.onPlayerChat(this, text);
```

minecrafter:
See #107 for my implementation

PR #107, opened 2014-06-07
Implement asynchronous chat support.

minecrafter:
This is like the support from Bukkit, which dictates asynchronous chat only for chat messages from the network.

This is my first contribution to Glowstone, so please don't bite hard.


zml2008:
This doesn't really take care of the point of async chat, which is being not tick-synchronized. You shouldn't have to worry about calling a scheduler task, just make sure player chat packets are handled on the async task executor in a non tick-synchronized way (since network threads are pooled, it's probably best to keep as much handling off of them as possible) and that the event is marked as async (if not it will be tick-synchronized).

Threading in GS works a bit differently than in CB thanks to using a thread-per-world model too, so watch out for that when you're working with stuff.

minecrafter:
Thanks for the clarification on how to handle asynchronous chat, @zml2008.

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -974,17 +978,28 @@ public void chat(String text) {
                 getServer().getLogger().log(Level.SEVERE, "Exception while executing command: " + text, ex);
             }
         } else {
-            // todo: async this
-            PlayerChatEvent event = EventFactory.onPlayerChat(this, text);
-            if (event.isCancelled()) {
-                return;
-            }
+            final String theText = text;
+            final boolean isAsync = async;
```
SpaceManiac: You can eliminate the need for these variables by marking the method's parameters final.

minecrafter: I did not know what convention you wanted, but I've fixed it now.

SpaceManiac:
Looks good on parameters. I'll investigate the sync/timing properties of the setup when I get the chance. However, it should be noted that the original, non-async PlayerChatEvent probably should still be fired (in sync) if there are any handlers registered for it (there should be a way to check this through `PlayerChatEvent.getHandlerList()` or whatever the method is named), in case old plugins are still trying to use it. CraftBukkit does this.

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -974,17 +978,40 @@ public void chat(String text) {
...
+                    }
+                }
+            };
 
-            String message = String.format(event.getFormat(), event.getPlayer().getDisplayName(), event.getMessage());
-            getServer().getLogger().info(message);
-            for (Player recipient : event.getRecipients()) {
-                recipient.sendMessage(message);
-            }
+            if (async)
```
turt2live: Not sure of the coding standards, but this looks kinda strange. In my opinion, it should be with brackets or only consume 2 lines.

minecrafter: It is strange, but in order to comply with the Bukkit API which mandates chat() to be synchronous and network chat to be asynchronous, this is the way I have to do it.

But since you did bring up the brackets, I'll add those.

turt2live: I meant it looks strange because of formatting, not for anything else :P

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -974,16 +978,40 @@ public void chat(String text) {
...
+                        recipient.sendMessage(message);
+                    }
+                }
+            };
 
-            String message = String.format(event.getFormat(), event.getPlayer().getDisplayName(), event.getMessage());
-            getServer().getLogger().info(message);
-            for (Player recipient : event.getRecipients()) {
-                recipient.sendMessage(message);
+            if (async) {
```
turt2live: GitHub really failed on this one....

src/main/java/net/glowstone/entity/GlowPlayer.java
minecrafter: Fixed with 2d4d6a7

src/main/java/net/glowstone/entity/GlowPlayer.java
dequis: What? You mean the way the diff was split?

turt2live:
@minecrafter This PR is not mergable. Please fix that as soon as you can. Thanks.

turt2live:
Related ticket: #103

dequis:
It conflicts mainly because of the newline changes, fix that with the `-X renormalize` merge strategy, so either `git rebase master -X renormalize` or `git merge master -X renormalize`. There's also a minor conflict in GlowScheduler because an `@Override` was added right after runAsynchronouslyImmediately.

Pushed two branches, one with the rebase I did:

https://github.com/dequis/Glowstone/tree/async-chat-rebase

And a squashed version because I love squashing things.

https://github.com/dequis/Glowstone/tree/async-chat-rebase-squash

dequis:
I tested this with craftirc3, which was the original reason I opened the previous ticket.

It just works™

minecrafter:
Merged with latest now.

turt2live:
Sorry - bot fail :(

src/main/java/net/glowstone/scheduler/GlowScheduler.java
```diff
@@ -262,6 +262,14 @@ public int scheduleAsyncRepeatingTask(Plugin plugin, Runnable task, long delay,
         }
     }
 
+    public Future<?> runTaskImmediateAsynchronously(Runnable task) {
```
JeromSar, line 265: This isn't necessary. `runTaskAsynchronously()` already handles running tasks asynchronously.

minecrafter, line 265: See https://github.com/GlowstoneMC/Glowstone/pull/107#issuecomment-45413730 for why this was implemented.

JeromSar, line 265: @minecrafter Alright, cool, thanks!

SpaceManiac:
I appreciate your help - based on your work, I took an alternate route in the implementation in d2e4454bc5804d9d237ec6a3524c61d8d072ce48.

PR #111, opened 2014-06-10
Remove org.bouncycastle libraries to reduce filesize significantly.

OverCaste:
They are unnecessary, as the javax libraries are faster and included in JDK implementations.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -89,7 +89,7 @@ repositories {
 // Dependencies used by our project
 dependencies {
     compile group: 'org.bukkit', name: 'bukkit', version: ext.bukkitVersion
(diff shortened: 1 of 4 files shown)


OverCaste:
Got ahead of myself there fixing the last bug. Will revert gradle and re-open.

SpaceManiac:
Think you could avoid rearranging imports too much? I know they're not especially well-organized at this point but it'd be nice not to make it worse until I decide on something better. If you want I can export my IDEA profile or something.

OverCaste:
I accidentally left import formats on for the project, it's already fixed, won't happen again.

SpaceManiac:
Testing proved successful. Merged in 98b9bdbc980bc42148f559f49ab957fad4a57118. Thanks!

PR #112, opened 2014-06-12
Change Item registry's hash map size from 16 to 512.

OverCaste:
This should increase speed when all of the items are added from Quadratic time to
Linear time.


kitskub:
But you increased it to 512...not 256 :P

SpaceManiac:
Merged in c8cba8b29e45dd1beb47ef726b24952ca5777fbf. Thanks!

Issue #98, opened 2014-06-07
Block drop mappings


dequis:
When breaking a grass block, you get a grass block in the inventory instead of dirt.
When breaking stone, you get stone instead of cobblestone
When breaking coal ore, you get the coal ore block instead of just the coal.
Etc. See: http://minecraft.gamepedia.com/Drops#Block_drops

Related: #4 (item pickup entities)

Needs handling for:
- [X] Generic `getDrops()` for Bukkit API.
- [x] Only dropping when correct tool is used.
- [ ] Dropping bonus loot for Fortune enchantments.
- [ ] Dropping the original block for Silk Touch enchantments.
- [ ] Dropping the correct items for blocks (1.8 blocks, for example)
  - Some kind of unit test should be done to test this however
  - Reference for bad mappings: #527

turt2live:
Updated description to include content from #218 (just added to checklist)

PR #120, opened 2014-06-17
Make blocks with non-standard drops drop the correct items

warriordog:
This makes blocks that do not drop normally drop the correct items when mined.  This includes blocks such as ice (drops nothing), ores (drop multiple items), crops, and more.  This adds Isssue #98.


warriordog:
This is why developing on windows is annoying, github thinks the file is BlockPackedICe.java, windows thinks it is BlockPackedIce.java...

SpaceManiac:
Hey, thanks for the contribution!

~~You should be able to correct the filename by using `git mv` to name the file to something completely different and then back to the name with the correct case.~~

On the style side, there's just a few things - for blocks with no drops or that always drop the same ItemStack (the ones where there's only a constructor that calls `setDrops`), it might make sense to make a `BlockDrops` class like currently exists for `ItemPlaceAs` so separate classes aren't needed for each thing. This reduces the amount of code and files needed but also puts some of the complexity inside `registerBuiltins` instead. Thoughts?

Also, overridden methods do not need to duplicate the javadoc unless it's changed somehow, and it might be worthwhile to make a helper method in `BlockType` that takes the `ItemStack...` and calls `Collections.unmodifiableList(Arrays.asList(...))` for you.

warriordog:
@SpaceManiac 

I fixed the wrong name, thanks anyway though!  I plan to rewrite a proper hierarchy for the blocks next.  I did it this way for now since many of those blocks have other specific functions that will need a unique class when they are added.

warriordog:
@SpaceManiac 
I added several BlockType subclasses that provide specific drop logic.  They have a constructor that accepts a Material, data value, and any numbers needed for whatever drop logic they contain.  I was able to remove most of the specific block classes this way.

I also removed the extra javadocs, that was something my IDE does automatically.

src/main/java/net/glowstone/block/blocktype/BlockTallGrass.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Random;
+
+public class BlockTallGrass extends BlockType {
+    private final Random random = new Random();
+
+    @Override
+    public Collection<ItemStack> getDrops(GlowBlock block) {
+        if (random.nextFloat() < 12.5) {
```
SpaceManiac: I think `nextFloat()` returns a value in the range [0, 1), so you'd want to check `< .125` instead.

warriordog: Oops your exactly right, I'll fix that!

SpaceManiac:
Alright, looking good, just one last thing: if you have the opportunity, completing that `//todo: add logic to drop correct sapling type and apples` in `BlockLeaves` would be much appreciated. Should be good to merge then.

Sleaker:
Does this drop logic work on a per-tool basis? This is how minecraft actually handles drops.  Blocks don't have static drops, it's all based on what tool you're using but the getDrops method seems to generate a static itemstack regardless of how the block is being destroyed.  This is not matching behaviour and the whole getDrops stuff probably needs to be made aware of the breaking tool.

Examples: Silk Touch allows non-standard blocks to drop.
Shears allow leaves to be dropped.
If you mine certain ores with a Pick that's not good enough nothing drops.

warriordog:
@Sleaker 

I believe that this is only called when broken by a fist, otherwise blocks would drop in creative mode and there would be a way to tell what the user is holding.  I believe there is a different area of the code for that, or else there is not yet support for tools.  Either way, the current block drop code does not provide a way to tell how the block was broken.

Sleaker:
looks like it's done in DiggingHandler. there are some problems with this design though, as block drops are as I mentioned based on various factors, not simply what type of block is bring broken.

warriordog:
@Sleaker 
Yeah it really should be combined into one method getDrops(GlowBlock block, BreadSource source) or something like that.

SpaceManiac:
The problem is a little tricky. The Bukkit API mandates that we expose the drops a block "should" have (as if they had been broken with the correct tool) through `Block#getDrops()`. The `BlockType#getDrops()` could be changed to take an ItemStack, or some more general system for tool types and levels could be set up to check for drops against specific tools if possible.

Originally I had planned to handle silk touch by bypassing `getDrops()` entirely and I think this still makes sense. Fortune and things like shears might be a little bit of design trouble, though.

warriordog:
@SpaceManiac 
Personally I think it would make more sense to use one method with an ItemStack parameter, but splitting it into different methods "getNormalDrops(Block block)", "getExplosionDrops(Block block, float explosionStrength)", getToolDrops(Block block, ItemStack tool)", etc. would make the code easier to read.  Vanilla minecraft does it this way, although the normal and tool drops are combined and the tool is null for a hand.

SpaceManiac:
Merged in 4c6cd4075676230bc32fb6a0625400bf65ee46d8. Thanks! Original ticket remaining open until the special drops stuff is solved.

Issue #124, opened 2014-06-19
tellraw command


deathcap:
The `tellraw` command is unimplemented:

```
>tellraw
13:15:36 [INFO] Unknown command "tellraw", try "help"
```

would be nice to have this :) usage is `/tellraw player raw-json-message`, allows sending formatted JSON text (spec at http://minecraft.gamepedia.com/Commands#Raw_JSON_Text) in the http://wiki.vg/Protocol#Chat_Message packet, beyond what the Bukkit API sendMessage() (and even sendRawMessage()) allows (preformatted into `{"text": …}`), until the RichMessage API at github.com/Bukkit/Bukkit/pull/1065 is implemented

Sleaker:
@deathcap - I don't know if that's a good idea, the reason why bukkit probably doesn't do it is because if you messup the json data you disconnect the client.  Just something to be aware of.

navarr:
It's probably not implemented in Bukkit because they'd have to do a validation layer before then.  Which honestly probably shouldn't be too difficult (other than the annoyance of parsing JSON)

SpaceManiac:
The command isn't in Bukkit proper because there's no Bukkit API to accomplish what it does. In CraftBukkit it's just a passthrough to the Mojang command.

This would be trivial to implement if error-checking was disregarded, but I'm not sure it's the best solution.

deathcap:
Well, a proper Bukkit API is probably the better solution, they're working on it at https://github.com/Bukkit/Bukkit/pull/1065 and it should solve the client validation problem since the server will generate the JSON message from the `RichMessage` API calls, which abstract away the JSON implementation details so it could refuse to generate invalid messages. Looks like this API is a ways off, though (but once/if it is integrated I'd be happy to switch to it).

Meanwhile, I think implementing the `/tellraw` command is at least, most expedient, and will increase plugin compatibility. I'm developing a plugin, testing on Glowstone, and couldn't find any compatible mechanism to send formatted text. It seems that plugin developers for other implementations either resort to calling into Minecraft to send the raw 0x02 http://wiki.vg/Protocol#Chat_Message packet, or using `dispatchCommand()` on the server to send `/tellraw`. I could chose the former but would have to implement it differently for each Bukkit server implementation; if Glowstone supports /tellraw then the same command would work on both Spigot and Glowstone.

[...]

PR #125, opened 2014-06-20
Add /tellraw command. Closes GH-124

deathcap:
https://github.com/SpaceManiac/Glowstone/issues/124


SpaceManiac:
`session` in `GlowPlayer` is exposed through `getSession`. I think I would prefer something like `glowPlayer.getSession().send(new ChatMessage...)` over creating a new method in GlowPlayer that's only used for one thing.

deathcap:
Changed to use glowPlayer().getSession() instead of a new GlowPlayer method (I almost did it this way initially)

SpaceManiac:
Merged in 1c8634692524fc87de44a601b8f98513cdc14a80. Thanks!

PR #133, opened 2014-08-25
Allow placement of blocks overriding tallgrass and fire.

justcool393:
Currently, players are unable to place inside the fire block, as you can in vanilla. This PR adds this ability.

Also, when a block is attempted to be replacing a tallgrass block, it creates the block on what the block was placed against.

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockType.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockType.java
@@ -117,8 +117,8 @@ public class BlockType extends ItemType {
         GlowBlock target = against.getRelative(face);
         GlowBlockState newState = target.getState();
 
-        // only allow placement inside tall-grass, air, or liquid
-        if (against.getType() == Material.LONG_GRASS) {
+        // only allow placement inside tall-grass, air, liquid or fire
+        if (against.getType() == Material.LONG_GRASS || against.getType() == Material.FIRE) {
             target = against;
         } else if (!target.isEmpty() && !target.isLiquid()) {
             //revert(player, target);
@@ -145,7 +145,7 @@ public class BlockType extends ItemType {
         newState.update(true);
 
         // play a sound effect
-        // todo: vary sound effect based on block type
+        // TODO: vary sound effect based on block type
         target.getWorld().playSound(target.getLocation(), Sound.DIG_WOOD, 1, 1);
 
         // do any after-place actions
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -117,10 +117,8 @@ public final void rightClickBlock(GlowPlayer player, GlowBlock against, BlockFac
         GlowBlock target = against.getRelative(face);
         GlowBlockState newState = target.getState();
 
-        // only allow placement inside tall-grass, air, or liquid
-        if (against.getType() == Material.LONG_GRASS) {
-            target = against;
-        } else if (!target.isEmpty() && !target.isLiquid()) {
+        // only allow placement inside tall-grass, air, liquids or fire
+        if (!target.isEmpty() && !target.isLiquid() && !(target.getType() == Material.FIRE || target.getType() == Material.LONG_GRASS)) {
```
ewized, line 121: I don't think this is the most efficient way of handing it.

SpaceManiac:
The change does seem to allow placement inside fire, but it also breaks placing blocks inside of tall grass by right-clicking directly on it. It's possible it didn't work before but the diff makes it look like you removed the check and it definitely doesn't work with this change.

It'd be great if you could get placement inside tall grass working again before I merged this.

kitskub:
@SpaceManiac the check is still there...

SpaceManiac:
When placing against tall grass, the server has to compensate by placing the block inside the block the client says it clicked against (the tall grass). That's what the removed "target = against" assignment is supposed to do. Honestly I think it was broken before but it'd be great to see it fixed here.

kitskub:
@SpaceManiac oh I see

justcool393:
Right now, trying to place over tall grass isn't possible to do at all. I'll look and see what I can do to fully fix it (it doesn't like me placing over it when I have it aimed at the tallgrass block.

turt2live:
what about snow layers?

justcool393:
Yeah, that will need to be implemented as well. Looks like it only replaces the block if the damage value is 0.

src/main/java/net/glowstone/block/blocktype/BlockType.java
turt2live, line 121: Sounds like a possible method in `Material` should be added.

turt2live:
@justcool393 this PR does not currently merge. Please fix that ASAP!

TomCaserta:
This is already fixed in the latest version in any case (including in fire). Overriding canOverride(block,face,holding) and canAbsorb(block,face,holding) methods in the BlockType file achieves the same results.

SpaceManiac:
A fix for this same issue was added in 19c7ec353a9a1c09b70e8ade5e8659c60ab13bfc. Thanks for the help!

Issue #147, opened 2014-09-04
Glowstone license does not comply with Bukkit license


Grinch:
~~As discussed in #glowstone, it has been determined that Glowstone's MIT license is not valid due to it implementing and bundling Bukkit. This means that Glowstone must become GPLv3 in order to comply with Bukkit's license.~~

Consensus has been: Glowstone's _distribution_ ("the software as it is actually run") must be licensed under the GPL. The source code can remain MIT. This ticket will be closed when the distribution includes a copy of the GPL.

Reference: http://www.gnu.org/licenses/gpl-faq.html#IfLibraryIsGPL

JeromSar:
@Grinch has a valid point. :)

Ribesg:
At first I was like "But MIT is GPL-Compatible", but Google says it only means you can use MIT things in GPL things, not the other way

gdude2002:
GPL makes me sad. It's like a frickin' headcrab.

dumptruckman:
Yeah... Just go GPLv3 and then don't take it out on plugins when they don't go GPLv3. :)

JeromSar:
@dumptruckman Bukkit is not included in plugins and thus these restrictions don't apply to them.

Ribesg:
Yeah but plugins can't exist at runtime without Bukkit

gdude2002:
It's a sticky and confusing situation, but it might be **safer** to assume that the GPL does apply to plugins unless someone can get a code lawyer to say otherwise.

Grinch:
The tricky part is that gnu.org states that it is included and therefore must comply, but it isn't included. Even if it doesn't work outside of the Bukkit environment it can be compiled and redistributed without ever including it in the same file.

JeromSar:
@Grinch is correct.

Zidane:
GPL is an infectious license basically this is how Glowstone/Bukkit developers become "legal"

Glowstone: License becomes GPL v3, same as Bukkit
Plugin Developers: License becomes GPL v3, same as Bukkit

There is no ifs ands or buts, the verbage is clear. If you use a GPL lib a part of your implementation then you must be GPL as well. Glowstone and Plugin Developers both MUST use this license to be in compliance.

Welcome to the "Open Source" movement.

JeromSar:
[...]

PR #148, opened 2014-09-04
Switched to GPLv3 license.

JeromSar:
See issue #147 For details.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,624 @@
 Glowstone Copyright (C) 2011-2014 Tad Hardesty.
 Lightstone Copyright (C) 2010-2011 Graham Edgecombe.
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
 
-The above copyright notice and this permission notice shall be included in
-all copies or substantial portions of the Software.
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
 
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-THE SOFTWARE.
+                            Preamble
 
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
+the GNU General Public License is intended to guarantee your freedom to
+share and change all versions of a program--to make sure it remains free
+software for all its users.  We, the Free Software Foundation, use the
+GNU General Public License for most of our software; it applies also to
+any other work released this way by its authors.  You can apply it to
+your programs, too.
+
+  When we speak of free software, we are referring to freedom, not
+price.  Our General Public Licenses are designed to make sure that you
+have the freedom to distribute copies of free software (and charge for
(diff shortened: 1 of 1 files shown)


Ribesg:
Need acceptance by the following people before being merged:
- [ ] @SpaceManiac 
- [ ] @grahamedgecombe 
- [ ] @zml2008 
- [x] @zhuowei 
- [ ] @desmin88 
- [x] @Grinch 
- [x] @akrieger 
- [ ] @hef 
- [x] @OverCaste 
- [ ] @lukespragg 
- [ ] @wtfaremyinitials 
- [ ] @kitskub 
- [ ] @deathcap 
- [ ] @warriordog 
- [ ] @UltraFTW 
- [ ] @ewized 

Just answer with "I agree" (or moar) here.

Grinch:
I hereby agree that all submitted code by myself, Steven Downer (aka @Grinch) to Glowstone is allowed to be relicensed to GPLv3.

gdude2002:
> Just answer with "I agree" (or moar) here.

Moar!

JeromSar:
The MIT license allows sub-licensing and is thus fully compatible with the GPLv3 license. No "consent" is required.

gdude2002:
I believe that we should still get consent to avoid kicking the contributors in the dick, so to speak, and doing something they haven't had a say in.

That said, if this takes more than a few days, we might have to do it. But it's not our project to do that to, let's wait for @SpaceManiac

OverCaste:
My patches are licensed under the "don't bother me unless it's for a good reason" license, which is fully compatible with GPL.

Count me in.

wtfaremyinitials:
@OverCaste is that a real license? I might need to start using that.—

basicer:
Please see http://www.gnu.org/licenses/gpl-faq.html#LinkingWithGPL

All that is required here is to make the binary JAR that is distributed on glowstone,net GPL.

dumptruckman:
I agree with @basicer 

This would allow a much friendlier licence (such as MIT) to be used for the source code.  (This means the PR needs to be fixed!)

dumptruckman:
Okay, apparently not so simple.
["Subclassing is creating a derivative work. Therefore, the terms of the GPL affect the whole program where you create a subclass of a GPL'ed class."](http://www.gnu.org/licenses/gpl-faq.html#OOPLang)

JeromSar:
@dumptruckman is correct, GlowStone must be licensed under GPLv3. However, no consent is required because the MIT license (current license) explicitly permits sub-licensing.

basicer:
@dumptruckman - Sub-classing still happens at build time, the "whole program" that is affected is the resulting JAR.

dumptruckman:
Changing to GPLv3 will make the project legal with no questions asked and doesn't appear to affect anything else.  The only case would be any further mods to Glowstone would be required to be GPLv3... like if Glowstone had its own spigot kinda thing.  Plugins are not affected by this change as they depend on Bukkit and their license issues appear to be an unrelated issue.

zhuowei:
I agree.

olliestanley:
Seems like changing to GPL is the best way, especially if it doesn't require consent.

basicer:
The fact that it doesn't require consent is a large part of the reason why nothing needs to be done.

Also, if people really want to make usage under the GPL explicit, why not instead move to a dual licensing system of MIT/GPL instead of removing the MIT license.  Removing the MIT license will removing freedoms people had to deal with the code that they previously had.

For example, NibbleArray is a pretty cool class.  Maybe I would like to use that in some unrelated Java project in the future.  Switching to the GPL license would mean I couldn't reuse that bit of code, even though it is totally unrelated to Bukkit.

For example, jQuery use to license their code under combination MIT/GPL license.  (They  later switched to using just the MIT license, because as I have been saying, having both is redundent http://blog.jquery.com/2012/09/10/jquery-licensing-changes/)

JeromSar:
@basicer The terms of GPL prevent dual-licencing. The original MIT license is not legally valid due to Bukkit being licensed under GPL.

basicer:
I'm not sure what your legal basis for that is.  Can you explain?  To me it seems unlikely a project as popular as jQuery would have had an "invalid license" for several years.

dumptruckman:
@basicer Craftbukkit was a popular project and had an invalid license for years. ;)

JeromSar:
@basicer The MIT license is not an invalid license. However, because this project includes Bukkt (GPL), GlowStone is legally required to be licensed under GPL too.

jQuery on the other hand completely owns their own code and thus may change licence whenever they like.

basicer:
http://www.gnu.org/licenses/gpl-faq.html#LinkingWithGPL says that a work (GlowStone) that is combined with a GPL Library (Bukkit) can be licensed under a GPL compatible license (MIT).  The resulting program (the JAR file) would then be licensed under the GPL.

That FAQ entry describes exactly what is happening here.  The glowstone sorce sitting alone in a GIT repository is not a derivative work of Bukkit.  When we compile and link GlowStone against Bukkit, we are creating a derivative work.

dumptruckman:
http://www.gnu.org/licenses/gpl-faq.html#OOPLang and http://www.gnu.org/licenses/gpl-faq.html#LinkingWithGPL seem to contradict each other for our specific case....

basicer:
Notice in that section it refers to the "whole program".  This is the JAR file.

akrieger:
I agree.

dumptruckman:
Hmm yeah, could be the case.  Someone should reach out to FSF and see if they can help.

SpaceManiac:
Closing this in favor of applying the GPL to only the distribution, not the source code, as discussed on IRC and elsewhere.

PR #149, opened 2014-09-04
Added Java 1.5 Override annotations where required

JeromSar:
This change adds the Override annotations where required and possible.

### DIFF
--- a/src/main/java/net/glowstone/ChunkManager.java
+++ b/src/main/java/net/glowstone/ChunkManager.java
@@ -326,10 +326,12 @@ public final class ChunkManager {
             cz = z;
         }
 
+        @Override
         public Biome getBiome(int x, int z) {
             return world.getBiome((cx << 4) | x, (cz << 4) | z);
         }
 
+        @Override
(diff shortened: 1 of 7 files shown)


coelho:
To be noted, in Java 1.5, you can not add Override annotations to implementations of\* interfaces.

JeromSar:
@coelho Correct, this behaviour changed in Java 1.6. We however compile in Java 7 so this shouldn't be a problem.

olliestanley:
I approve of this, I hate reading through code where there are some methods which are overriden and some which aren't but you can't tell which is which because there is no annotation.

SpaceManiac:
I will double-check nothing got missed, merge this, then include it in the style guide when that gets written.

JeromSar:
@SpaceManiac I set my compiler to fail compiling if Override annotations weren't present. Then I auto-fixed all missing annotations automatically. Gosh, NetBeans is awesome. :)

SpaceManiac:
Fixed manually in a357d2c62661ac89910688e499e849f026c9e7dc. Thanks!

Issue #150, opened 2014-09-04
Clients can connect more than once


ghost:
Yesterday while testing Glowstone I found out this issue, I'm not sure how to replicate it (or if it can actually be replicated) but apparently trying to connect with my minecraft client ended up in connecting three times for some reason

```
2014/09/03 23:06:21 [INFORMACIÓN] Konato_K [/127.0.0.1:54589] connected, UUID: 11697c28-9c63-31ba-8f90-07e3c49c59d2
2014/09/03 23:06:21 [INFORMACIÓN] §eKonato_K joined the game
2014/09/03 23:06:23 [INFORMACIÓN] Konato_K [/127.0.0.1:54590] connected, UUID: 11697c28-9c63-31ba-8f90-07e3c49c59d2
2014/09/03 23:06:23 [INFORMACIÓN] §eKonato_K joined the game
2014/09/03 23:06:25 [INFORMACIÓN] Konato_K [/127.0.0.1:54591] connected, UUID: 11697c28-9c63-31ba-8f90-07e3c49c59d2
2014/09/03 23:06:26 [INFORMACIÓN] §eKonato_K joined the game
```

This appeared on the chat and on the client it looked like this
![Image](http://i.imgur.com/9Qrz8A1.png)

As you can see there are 3 of me, one being me (the camera) another being a player that never moved, and the other being another copy of me that followed what I was doing (with a slightly delay, that's why you can see it's head), this happened with build 46, but I didn't found any changes about this for 47 so I think it still may be valid, as a note I only opened one client, but for some reason it connected 3 times, if you want to see the full log it's [here](http://pastebin.com/VgbAQhTy)

JeromSar:
Looking into this.

JeromSar:
I believe I fixed the problem but I cannot test due to #145.  When the issue is resolve, I'll hopefully submit a PR with a patch. :)

SpaceManiac:
Was this produced with a single MC client open, and a single connection to the server made? Did you cancel while connecting and then connect again, or anything like that?

ghost:
It was a single client, I didn't open more than one, and I tried to connect after disconnecting (as the full log shows) and for some reason it logged three times, I didn't cancel the connection.

Issue #28, opened 2011-07-20
The Big List Of Events


SpaceManiac:
A list of all events and in what manner they are implemented in Glowstone. Events checked off here are emitted by Glowstone in what is believed to be all appropriate cases; other issues may document incompleteness or errors in the implementation of these events.

block:
- [x] BlockBreakEvent
- [ ] BlockBurnEvent
- [x] BlockCanBuildEvent
- [x] BlockDamageEvent
- [ ] BlockDispenseEvent
- [ ] ~~BlockEvent~~
- [ ] BlockExpEvent
- [ ] BlockFadeEvent
- [ ] BlockFormEvent
- [ ] BlockFromToEvent
- [ ] BlockGrowEvent
- [ ] BlockIgniteEvent
- [ ] BlockMultiPlaceEvent
- [ ] BlockPhysicsEvent
- [ ] ~~BlockPistonEvent~~
- [ ] BlockPistonExtendEvent
- [ ] BlockPistonRetractEvent
- [x] BlockPlaceEvent
- [ ] BlockRedstoneEvent
- [ ] BlockSpreadEvent
- [ ] EntityBlockFormEvent
- [ ] LeavesDecayEvent
- [ ] NotePlayEvent
- [ ] SignChangeEvent

enchant:
- [ ] EnchantItemEvent
- [ ] PrepareItemEnchantEvent

entity:
- [ ] CreatureSpawnEvent
- [ ] CreeperPowerEvent
- [ ] EntityBreakDoorEvent
- [ ] EntityChangeBlockEvent
- [ ] EntityCombustByBlockEvent
- [ ] EntityCombustByEntityEvent
- [ ] EntityCombustEvent
- [ ] EntityCreatePortalEvent
- [ ] EntityDamageByBlockEvent
- [ ] EntityDamageByEntityEvent
- [ ] EntityDamageEvent
- [ ] EntityDeathEvent
- [ ] ~~EntityEvent~~
- [ ] EntityExplodeEvent
- [ ] EntityInteractEvent
- [ ] EntityPortalEnterEvent
- [ ] EntityPortalEvent
- [ ] EntityPortalExitEvent
- [ ] EntityRegainHealthEvent
- [ ] EntityShootBowEvent
- [ ] EntityTameEvent
- [ ] EntityTargetEvent
- [ ] EntityTargetLivingEntityEvent
- [ ] EntityTeleportEvent
- [ ] EntityUnleashEvent
[...]

PR #153, opened 2014-09-04
Revamp login logic and fix multiple players being able to log in.

JeromSar:
This commit modifies LoginHandlerStart to check for existing players before allowing the client to login.


RhvsW:
Default behaviour in vanilla is to kick the other player from the server and allow the new one to join.

JeromSar:
@MeltedLux I believe it's not Glowstone's goal to mimic Vanilla. It fails to do so on several fronts. However, if requested, I'll re-implement the patch to correct this. :)

RhvsW:
Well it would be nice to stick closely to it.

olliestanley:
I think it should stick with what you've done @JeromSar - if people share accounts they can constantly kick each other off and that's just ridiculous IMO.

SpaceManiac:
There is some use to kicking off the existing player in case of connection problems, so you don't need to wait for your old connection to timeout before you can rejoin.

Shared accounts is handled by the way auth is done - once the first player is kicked off, they cannot log in to the server again until they restart the entire game.

JeromSar:
Alright, switched to kicking the current online player instead. Tested and well. :)

Ribesg:
You're supposed to be able to prevent same person to connect by kicking the connecting one, in `PlayerLoginEvent`'s Result, I think. Does it handles that currently?

Maybe some timeout

JeromSar:
You're right @Ribesg. I've also noticed the deprecated (but still valid) PlayerPreLoginEvent is not implemented. I'll implement that in a separate pull request.

SpaceManiac:
The `bungee` branch includes an implementation of Async/PlayerPreLoginEvent.

SpaceManiac:
Implementation looks solid. I'll give it a test and merge as soon as I can.

turt2live:
This doesn't work. It kicks the player that tried to log in if they were kicked due to a server error previously. Basically it prevents them from logging in.

JeromSar:
@turt2live Huh? The code works as expected.

turt2live:
@JeromSar I'll create a test case in a bit.

turt2live:
@JeromSar I just pulled this and compiled it locally. Once running, I can get onto the server (it's just me) get a good portion of the world, and then I'm booted off. 

Something is wrong with the PR.

SpaceManiac:
> Once running, I can get onto the server (it's just me) get a good portion of the world, and then I'm booted off.

This is covered in a related ticket, #156.

turt2live:
The ability to see the world is, but the PR kicks the connecting player off the server as it somehow detects the player as a duplicate.

SpaceManiac:
Oh, I see. It boots the player for having the same UUID as... themselves. I guess it should either handle that case or just involve some fix to #156 which doesn't add the player to the world until all these checks have passed.

Also, what about the case where the `PlayerKickEvent` is cancelled? Should this be allowed? What does CB do? Sort of a corner but it's good to think about these things.

JeromSar:
Writing this from scratch, going to revamp the login logic a bit. Also implementing the missing login events proper. Might take some time.

SpaceManiac:
Check out commit 682b9c5147f5fd6f9dc992163a3b587ab6d8a97d on the `bungee` branch for some clues regarding the missing login events.

EntityReborn:
This causes a merge conflict.

JeromSar:
@EntityReborn I saw, fixing now.

JeromSar:
@SpaceManiac Addressing your question about the PlayerKickEvent. I took a peek at how CraftBukkit handles it, and it seems the PlayerKickEvent is never fired when kicking a player that is already online. I followed suit. :)

EntityReborn:
What? Yes it is! We have code in CommandHelper that hooks this event. Pretty sure it does get fired.
EDIT: Just verified with a CH user, yes it does get fired.

src/main/java/net/glowstone/EventFactory.java
```diff
@@ -131,8 +133,22 @@ public static PlayerLoginEvent onPlayerLogin(GlowPlayer player) {
...
     }
 
-    public static PlayerPreLoginEvent onPlayerPreLogin(String name, GlowSession session) {
-        return callEvent(new PlayerPreLoginEvent(name, session.getAddress().getAddress()));
+    public static AsyncPlayerPreLoginEvent onPlayerPreLogin(String name, InetSocketAddress address, UUID uuid) {
+        final AsyncPlayerPreLoginEvent asyncEvent = new AsyncPlayerPreLoginEvent(name, address.getAddress(), uuid);
+        callEvent(asyncEvent);
+
+        if (PlayerPreLoginEvent.getHandlerList().getRegisteredListeners().length > 0) {
+
```
turt2live: Whitespace

src/main/java/net/glowstone/net/handler/login/EncryptionKeyResponseHandler.java
```diff
@@ -105,19 +111,19 @@ public void handle(GlowSession session, EncryptionKeyResponseMessage message) {
...
             this.username = username;
             this.postURL = BASE_URL + "?username=" + username + "&serverId=" + hash;
-            setName("ClientAuthThread{" + username + "}");
         }
 
         @Override
         public void run() {
             try {
-                URLConnection conn = new URL(postURL).openConnection();
+                // cuthenticate
```
turt2live: ... psellign

src/main/java/net/glowstone/net/handler/login/EncryptionKeyResponseHandler.java
```diff
@@ -128,20 +134,20 @@ public void run() {
                 }
 
                 final String name = (String) json.get("name");
-                final String id = (String) json.get("id");
-                final JSONArray propsArray = (JSONArray) json.get("properties");
 
-                // Parse UUID
+                // parse UUID
```
turt2live, line 138: Unrequired change

JeromSar, line 138: @turt2live Judging from the other documentation in this project, the standard is to begin with a lowercase. The change is not required, but it's not unhelpful either. :)

src/main/java/net/glowstone/EventFactory.java
```diff
@@ -1,10 +1,10 @@
 package net.glowstone;
 
+import java.net.InetSocketAddress;
```
turt2live, line 3: Wrong import section

src/main/java/net/glowstone/net/handler/login/EncryptionKeyResponseHandler.java
turt2live, line 138: There is no standard defined, but right now it's just an unrequired change. @SpaceManiac - Should comments be enforced in any particular way?

SpaceManiac, line 138: All I'm looking for right now is a space between `//` and text if it's a textual comment, for readability. No worries on caps vs. lowercase, either way is acceptable.

turt2live, line 138: Ok then :3

turt2live:
Ticket: #150

src/main/java/net/glowstone/EventFactory.java
JeromSar, line 3: Fixed.

turt2live:
@JeromSar This is currently not mergable. Please fix that ASAP!

JeromSar:
@turt2live That would be the third time so. :/ Doing tonight.

JeromSar:
@turt2live Should be able to merge now. Formatting concerns have been addressed.

JeromSar:
Relative parent #28

TehKittyCat:
In fixing this, it would be good to fix the basis of #156 -- the world being shown before kick. The problem is GlowSession creates the GlowPlayer, which has no logic to wait for the result of the login events before sending the client LoginSuccess, JoinGame, the initial chunks and everything else.

src/main/java/net/glowstone/net/handler/login/EncryptionKeyResponseHandler.java
```diff
@@ -94,10 +97,12 @@ public void handle(GlowSession session, EncryptionKeyResponseMessage message) {
...
 
         // start auth thread
-        new ClientAuthThread(session, session.getVerifyUsername(), hash).start();
+        Thread clientAuthThread = new Thread(new ClientAuthRunnable(session, session.getVerifyUsername(), hash));
+        clientAuthThread.setName("ClientAuthThread{" + session.getVerifyUsername() + "}");
+        clientAuthThread.start();
     }
 
-    private static class ClientAuthThread extends Thread {
+    private static class ClientAuthRunnable implements Runnable {
```
SpaceManiac, line 105: Curious about the reason for this change.

JeromSar, line 105: @SpaceManiac Implementing a Runnable instead of extending a Thread is philosophically more correct. When you extend an object, you're implying you'll change it's behaviour. `ClientAuthThread`doesn't modify any existing behavior in the Thread object (doing so might be dangerous too!). Instead, you're only giving a thread something to run, a Runnable.

http://stackoverflow.com/questions/541487/implements-runnable-vs-extends-thread
http://howtodoinjava.com/2013/03/12/difference-between-implements-runnable-and-extends-thread-in-java/

SpaceManiac:
Merged in 266ba97c1c910227aef6f0a8536d430d5c9aa6d1.

Issue #77, opened 2014-02-11
ItemMeta implementations


SpaceManiac:
There is currently a generic ItemMeta, `GlowMetaItem`, but no specific implementations for the different ItemMeta subinterfaces. These should be added and implemented:
- [x] BookMeta
- [ ] EnchantmentStorageMeta (re-use normal enchants map?)
- [x] FireworkEffectMeta
- [x] LeatherArmorMeta
- [ ] MapMeta
- [ ] PotionMeta
- [x] SkullMeta
- [x] BannerMeta (added post-Bukkit)

Also, these need to be working for everything (sort of complicated, maybe look at CraftBukkit):
- [ ] Bukkit serialization and deserialization
- [ ] NBT serialization and deserialization
- [ ] Equality comparison (and hashCode stuff)
- [ ] Cloning (pretty simple)

ZephireNZ:
@SpaceManiac Skull and Banner meta have been implemented, so can be crossed off the list.

turt2live:
Thanks @ZephireNZ

ZephireNZ:
Currently working on proper serialization and comparing.

PR #158, opened 2014-09-05
Implemented GlowMetaLeatherArmor partially completes #77

TomCaserta:
Hey, 

(See #77)

Just getting into the whole open source thing, thought id start off small. Please let me know if I have made any mistakes.

Thanks,


TomCaserta:
Wait, forgot to change the default color whilst testing.

EntityReborn:
Just a suggestion of professionalism, when originally submitting your PR, keep the commit count down, and use sensible, informant, professional commit messages. "What is this even doing now" does not describe at all what that commit contains. Also, the git ignore commits could have been contained in other commits. Just a thought :)

TomCaserta:
Sorry, was using the gui git and couldn't get it to revert without it deleting everything, had to make that commit before it would do it correctly. Now that I know how to do it properly without messing it up it shouldn't be an issue in the future.

src/main/java/net/glowstone/inventory/GlowItemFactory.java
```diff
@@ -13,7 +13,7 @@
 public final class GlowItemFactory implements ItemFactory {
 
     private static final GlowItemFactory instance = new GlowItemFactory();
-    private static final Color LEATHER_COLOR = Color.fromRGB(0xA06540);
+    static final Color LEATHER_COLOR = Color.fromRGB(0xA06540);
```
Zidane, line 16: Why is this in this class...?

(Read comment below)

src/main/java/net/glowstone/inventory/GlowMetaLeatherArmor.java
```diff
@@ -0,0 +1,83 @@
...
+import org.bukkit.Material;
+import org.bukkit.Color;
+import org.bukkit.inventory.meta.LeatherArmorMeta;
+
+
+import java.util.Map;
+
+class GlowMetaLeatherArmor extends GlowMetaItem implements LeatherArmorMeta {
+    private Color color = LEATHER_COLOR;
+
```
Zidane, line 16: Put the static constant for the default color here

src/main/java/net/glowstone/inventory/GlowItemFactory.java
SpaceManiac, line 16: Bukkit API requires `getDefaultLeatherColor()` on `ItemFactory`, so `GlowItemFactory` is where this constant is kept.

Making the constant package-private instead of private is a little odd but not the worst. `GlowItemFactory.instance().getDefaultLeatherColor()` would also suffice.

PR #160, opened 2014-09-06
Replace Bukkit with Spigot-API

deathcap:


ewized:
Why?

EntityReborn:
What? No.

lukespragg:
Replacing one zombified API for another doesn't really solve anything. As @SpaceManiac said, best to wait to see what happens once the dust settles and go from there. If anything, it'd be best to maintain a fork of Bukkit directly, and pull in any specific fixes from other forks if needed/wanted.

SpaceManiac:
I'd need to see some pretty strong justification to accept this.

EntityReborn:
And the current drama with the DMCA's doesn't count :)

lukespragg:
It actually does @EntityReborn, as the fate of the APIs may change based on if they are updated by Mojang/Dinnerbone or not. So that's why it'd be best to wait and see what happens, and go from there. No sense jumping to a Bukkit fork or forking it directly until absolutely necessary.

deathcap:
Alright I'll close this PR, seems its not the way to go.

PR #175, opened 2014-09-06
Replace ConcurrentHashMap with CopyOnWriteArrayList

HiddenCloud:
A ConcurrentHashMap was unnecessary used to save the sessions, using a list is much faster.

### DIFF
--- a/src/main/java/net/glowstone/net/SessionRegistry.java
+++ b/src/main/java/net/glowstone/net/SessionRegistry.java
@@ -1,7 +1,6 @@
 package net.glowstone.net;
 
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 /**
  * A list of all the sessions which provides a convenient {@link #pulse()}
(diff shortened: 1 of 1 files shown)


Zidane:
While at first glance this seems like a solid change, this isn't much better if you ever expect Glowstone to have hundreds of players on at once.

With each addition to your list, java copies the list as a new list and adds your element. Unless you are reading roughly 90% of the time from the session registry, this will be a big performance detriment.

Again it depends, if you have a large server with 1000+ players and people constantly logging in and out....this will probably choke.

Also your assumption on a list being "much faster" is a fallacy; it depends on the hashcode of the object being stored in ConcurrentHashMap. If your object has a slow equals method, your list will still be even slower than it can be.

SpaceManiac:
Not convinced this is beneficial, as argued by Zidane above. I'd need to see some stronger justification based on what is currently happening under the hood and what this change brings (possibly backed by profiling?) to accept this.

To be fair, except under heavy join/depart load, the session list is indeed read much more often than it is written, but it is still written fairly frequently.

HiddenCloud:
Ah alright, I'll just leave it be then for now.

PR #178, opened 2014-09-07
Skylight in empty chunks should be 15, versus 0

mikeprimm:
Simple tweak encountered while working on Dynmap for Glowstone : block skylight values in empty chunks should be 15, versus 0 (this matches Bukkit behavior, and is logical for an all-air model for an empty chunk)


RhvsW:
:+1:

SpaceManiac:
Accepted - will be merging tomorrow.

mikeprimm:
Two additional updates - one more case of the skylight for empty case (for empty sections), and a fix for the isSectionEmpty() method, which has the opposite logic to what is needed (it returns true for non-empty sections now).

SpaceManiac:
Merged in 09062c0d7379354f94a26597229d1e26babfa8e3.

PR #182, opened 2014-09-07
Added: Hoppers

### LINKED ISSUE #119: Block placement (stairs, etc.)
BlockTypes representing stairs, slabs, logs, leaves, ladders, and other assorted blocks with facing need to be created to allow these to be placed correctly. The `placeBlock` function should be overridden to modify the passed BlockState to set data based on the player's facing or the face placed against as needed.

Blocks involved:
- [x] Torches
- [ ] Redstone torches
- [x] Logs
- [ ] Leaves
- [x] Slabs
- [ ] Tall flowers
- [ ] Pistons
- [x] Stairs
- [X] Sign posts and wall signs
- [x] Doors
- [ ] Rails
- [x] Ladders
- [x] Furnaces
- [x] Chests / Ender chests
- [x] Dispensers
- [x] Droppers
- [x] Hoppers
- [x] Levers
- [x] Buttons
- [ ] Pumpkins / Jack o'Lanterns
- [ ] Redstone repeaters / comparators
- [x] Trap doors
- [x] Vines
- [x] Fence gates
- [ ] Cocoa beans
- [ ] Tripwire hooks
- [x] Heads
- [x] Hay
- [x] Quartz
- [x] Beds
- [ ] Anvils
- [x] Banners

Edit (turt2live): Issue scope expanded to general placement problems including data not being set (leaves, etc) and not just facing directions.

anna-is-cute:
Partially fixes #119. Again, no functionality, only placement fixes.

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockHopper.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockHopper.java
@@ -1,0 +1,59 @@
+package net.glowstone.block.blocktype;
+
+import net.glowstone.GlowChunk;
+import net.glowstone.block.GlowBlockState;
+import net.glowstone.block.entity.TEHopper;
+import net.glowstone.block.entity.TileEntity;
+import net.glowstone.entity.GlowPlayer;
+import org.bukkit.Material;
+import org.bukkit.block.BlockFace;
+import org.bukkit.block.BlockState;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.util.Vector;
+
+public class BlockHopper extends BlockContainer {
+
+    public BlockHopper() {
+        setDrops(new ItemStack(Material.HOPPER));
+    }
+
+    public void setFacingDirection(final BlockState bs, final BlockFace face) {
+        byte data;
+        switch (face) {
+            case DOWN:
+                data = (byte) 0;
+                break;
+            case UP:
+                data = (byte) 1;
+                break;
+            case NORTH:
+                data = (byte) 2;
+                break;
+            case SOUTH:
+                data = (byte) 3;
+                break;
+            case WEST:
+                data = (byte) 4;
+                break;
+            case EAST:
+                data = (byte) 5;
+                break;
+            default:
+                data = (byte) 5;
+                break;
+        }
+        bs.setRawData(data);
+    }
+
+    @Override
+    public TileEntity createTileEntity(GlowChunk chunk, int cx, int cy, int cz) {
+        return new TEHopper(chunk.getBlock(cx, cy, cz));
+    }
+
+    @Override
+    public void placeBlock(GlowPlayer player, GlowBlockState state, BlockFace face, ItemStack holding, Vector clickedLoc) {
+        super.placeBlock(player, state, face, holding, clickedLoc);
+        setFacingDirection(state, face.getOppositeFace());
+    }
+
+}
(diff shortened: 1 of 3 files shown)


Tonodus:
Why did you override getDrops()?
Seems like the default behaviour without setting the data to the ItemStack. Is that necessary?

anna-is-cute:
@Tonodus If you want to get a no-texture Hopper ItemStack with data on it, then you should remove the getDrops(). ;)

Tonodus:
Why not? ;) No, seriously, is the hopper a special case or do you need to do this with every BlockType that has metadata for directional placing, etc.?

anna-is-cute:
It seems as if you must do this for each block that has directional placing. Can verify with levers, at least.

On September 7, 2014 at 1:45:47 PM, Tonodus (notifications@github.com) wrote:

Why not? ;) No, seriously, is the hopper a special case or do you need to do this with every BlockType that has metadata for directional placing, etc.?

JeromSar:
@jkcclemens You could use `setDrops()` in the constructor.

anna-is-cute:
I could.

On September 7, 2014 at 2:28:10 PM, Jerom van der Sar (notifications@github.com) wrote:

@jkcclemens You could use setDrops() in the constructor.

JeromSar:
@jkcclemens You could.

SpaceManiac:
For situations which allow it, I think prefer setDrops in constructor rather than overriding of getDrops. This allows for Find Usages-based searches to distinguish between the two types of drop handling.

src/main/java/net/glowstone/block/entity/TEHopper.java
```diff
@@ -0,0 +1,12 @@
...
+
+import net.glowstone.block.GlowBlock;
+import org.bukkit.event.inventory.InventoryType;
+
+public class TEHopper extends TEContainer {
+
+    public TEHopper(GlowBlock block) {
+        super(block, InventoryType.HOPPER);
+        setSaveId("Hopper");
+    }
```
turt2live, line 11: Your other files have a blank line after this - is this intentional?

src/main/java/net/glowstone/block/blocktype/BlockHopper.java
```diff
@@ -0,0 +1,59 @@
...
+
+public class BlockHopper extends BlockContainer {
+
+    public BlockHopper() {
+        setDrops(new ItemStack(Material.HOPPER));
+    }
+
+    public void setFacingDirection(final BlockState bs, final BlockFace face) {
+        byte data;
+        switch (face) {
```
turt2live, line 22: I'm not sure of the coding standards for this, but this may look better (and shorter) if on one line.

```
switch(face){
    case UP: data = (byte) 0; break;
    case DOWN: data = (byte) 1; break;
    // ...
}
```

anna-is-cute, line 22: Personally, I'm not a fan of compacting it so much. Given the contribution standard of adding braces to one-line if statements, I figure that expanding this switch-case is the right course of action.

SpaceManiac:
Merged in 8d9cf2f.

src/main/java/net/glowstone/block/blocktype/BlockHopper.java
turt2live, line 22: Fair point. I have no complaints :P

PR #185, opened 2014-09-07
Replace Bukkit API dependency with MIT-licensed alternative

deathcap:
(Not ready to pull, opening this PR for discussion/review to see if it is worth developing further.) Replaces the Bukkit API dependency (GPLv3) with a [clean room MIT-licensed alternative](https://github.com/deathcap/BedrockAPI) written using the API documentation javadoc available at http://jd.bukkit.org/rb/apidocs/ for reference.

Relates to #148 #147 in that this would allow licensing the Glowstone binary distribution more permissively than GPL (assuming there are no other relevant dependencies).

Currently Glowstone with this change _compiles_ successfully (e.g., all the needed interfaces are present) but is not yet runnable, due to missing method implementations. However I suspect is feasible to reimplement these missing pieces, if there is interest — though primarily an API, "Bukkit" also includes some important utilities in org.bukkit.*:
- [ ] plugin loader (plugin.yml parsing, class loading)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -51,7 +51,7 @@ mainClassName = "net.glowstone.GlowServer"
 
 // Extended project information
 ext.url = 'https://github.com/SpaceManiac/Glowstone'
-ext.bukkitVersion = '1.7.9-R0.1'
+ext.bukkitVersion = '0.0.1'
 ext.gitDescribe = getVersionName()
 
 // Minimum Java version required
@@ -88,7 +88,7 @@ repositories {
 
 // Dependencies used by our project
 dependencies {
-    compile group: 'org.bukkit', name: 'bukkit', version: project.ext.bukkitVersion
+    compile group: 'com.github.deathcap', name: 'bedrock-api', version: project.ext.bukkitVersion
     compile 'com.flowpowered:flow-networking:1.0.0-SNAPSHOT'
     compile 'jline:jline:2.11'
     //compile 'mysql:mysql-connector-java:5.1.14'
--- a/src/main/java/net/glowstone/ConsoleManager.java
+++ b/src/main/java/net/glowstone/ConsoleManager.java
@@ -125,7 +125,7 @@ public final class ConsoleManager {
     }
 
     private String colorize(String string) {
-        if (string.indexOf(ChatColor.COLOR_CHAR) < 0) {
+        if (string.indexOf(ChatColor.COLOR_CHAR.ordinal()) < 0) {
             return string;  // no colors in the message
         } else if (!jLine || !reader.getTerminal().isAnsiSupported()) {
             return ChatColor.stripColor(string);  // color not supported
--- a/src/main/java/net/glowstone/EventFactory.java
+++ b/src/main/java/net/glowstone/EventFactory.java
@@ -123,7 +123,7 @@ public final class EventFactory {
             event.disallow(PlayerLoginEvent.Result.KICK_BANNED, "Banned: " + ipBans.getBanEntry(address).getReason());
         } else if (server.hasWhitelist() && !player.isWhitelisted()) {
             event.disallow(PlayerLoginEvent.Result.KICK_WHITELIST, "You are not whitelisted on this server.");
-        } else if (server.getOnlinePlayers().length >= server.getMaxPlayers()) {
+        } else if (server.getOnlinePlayers().size() >= server.getMaxPlayers()) {
             event.disallow(PlayerLoginEvent.Result.KICK_FULL,
                     "The server is full (" + player.getServer().getMaxPlayers() + " players).");
         }
--- a/src/main/java/net/glowstone/GlowOfflinePlayer.java
+++ b/src/main/java/net/glowstone/GlowOfflinePlayer.java
@@ -16,7 +16,7 @@ import java.util.UUID;
 /**
  * Represents a player which is not connected to the server.
  */
-@SerializableAs("Player")
+//@SerializableAs("Player") // TODO: fix cannot find value()
 public final class GlowOfflinePlayer implements OfflinePlayer {
 
     private final GlowServer server;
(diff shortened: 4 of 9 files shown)


turt2live:
This will never fly. You have taken the Bukkit API source and committed it elsewhere. That is not only stupid, but incredibly rude. It's somewhat sickening to see that (not to mention you've removed MY work and claimed it as YOUR work, not to mention the other contributors).

deathcap:
@turt2live It's not using any original Bukkit API source, only based on for reference the documentation at http://jd.bukkit.org/rb/apidocs/. If this is not kosher, how else could one acceptably clean-room this dependency?

turt2live:
I can SEE the mistakes I made way back when I committed the changes in your source. You copied and pasted the source and removed some documenting attributes. To be completely free of any legal standing (because by the way, what you've done is illegal) you need to write it all yourself under the license you wish to hold it. You have not done that because 1) it's impossible in the timespan given and 2) I can clearly see you have not done so.

Try again.

deathcap:
@turt2live The Bukkit source repository was not used here, everything is only based on the API documentation reference. If a GPL source similarity made it in then it must have been additionally part of the documentation since that is all I looked at (or otherwise coincidental original work, specifics needed). This includes the class, method, and field names required for binary compatibility purposes.

turt2live:
You stole code and are using the wrong terms in your statements. You might as well just say it.

deathcap:
@turt2live I haven't seen your code, only the published API documentation. 

---

Regarding Travis CI test below, it is failing due to missing implementation (stubbed out APIs). For testing, skipping with `./gradlew -x test`. The resulting binary fails to run (for the same reason):

```
SEVERE: Error during server startup.
java.lang.NullPointerException
    at net.glowstone.util.ServerConfig.<init>(ServerConfig.java:63)
    at net.glowstone.GlowServer.parseArguments(GlowServer.java:172)
    at net.glowstone.GlowServer.main(GlowServer.java:80)
```

may or may not fix this.. leaning towards not, if the legality of reimplementing the Bukkit API from javadoc is called into question. Leaving this PR open for now but might close it depending (well, at least got this neat little tool out of this effort: https://github.com/deathcap/unjavadoc - potentially useful for other projects).

turt2live:
If you only published the API documentation, why is my code in your repo without credit?

deathcap:
@turt2live Haven't read your code, cannot comment on any alleged resemblance. The goal is to only match the required interfaces per the API specification so Bukkit plugins will be binary compatible, any implementation is rewritten from scratch (or non-existent for now, 99% is currently stubs). If I messed this up let's discuss further on IRC, shoot me a pm, would be happy to clear it up.

JeromSar:
@deathcap There's several issues with this pull request.
- The Bukkit alternative you specified fails to implement some _required_ implementation such as `Bukkit` and `PluginLoader`. Glowstone won't load plugins properly without these.
- Additionally, this code is without doubt based on the original Bukkit API as @turt2live suggested. Therefore, it applies to the GPL restrictions.
- The API isn't future proof (it'll need expanding in future versions). This is too much for GlowStone's developers to currently handle.
- We don't need a MIT-compliant dependency. See #148.

Overall, I don't think this is the correct way to go.

turt2live:
The discussion can continue here, where it makes more logical sense.

What you have done is taken the Bukkit source, stripped some code (which is not hard), threw it on a repo, and applied a license to it. Currently, Glowstone depends on a non-stripped Bukkit API so it can operate. 

By simply removing code from methods and removing some headers that state author information you have done nothing by violate laws and copyright. Putting them under a new name on a new repo with a new license is just adding whipped cream and a cherry. 

Quite honestly, this PR is illegal to even exist and even if it was legal it would never be pulled because it cannot function. You are essentially submitting a "get into jail free" card with some broken API. I strongly suggest you close your PR and pretend you never tried to submit it.

SonarBeserk:
I think that closing this pr would be a good idea.

RhvsW:
Yeah this is just plain stupid.

basicer:
@turt2live - Can you point to a place in @deathcap 's code the code where your code has been used?  There doesn't even seem to be any code from Bukkit in there, as nothing works.  Even things like Bukkit#getServerName return null instead of a sensible value.

If your argument is just that the code contains the same method names and signatures (including some deprecated ones), its is the view of the Free Software Foundation that merely copying these "header files" does not constitute a derivative work.

http://lkml.iu.edu//hypermail/linux/kernel/0301.1/0362.html

turt2live:
@basicer Taking the source files and stripping them to only the signatures is illegal. You might as well go to a car dealership and remove the seats from a car and drive off with it.

basicer:
@turt2live - Do you have a legal precedent supporting that?  I think its similar to going to a car dealer ship, looking at a car, and collecting a list of requirements that you would need to create a car that would fulfill all of the requirements your car has to pass to drive on the highway, then creating your own car.

JeromSar:
@basicer I don't think it matters, this PR does not compile and won't likely be accepted. @turt2live

basicer:
@deathcap said that it doesn't compile yet, but I imagine he intends to run around and fill in the method stubs with working code he will write himself.

SpaceManiac:
While I find the unjavadoc tool neat and appreciate your efforts, with Glowstone's own license "problem" resolved and Sponge on the horizon, I don't believe this is the way to move forward.

deathcap:
@MeltedLux Yeah this was the worst idea ever :/. Withdrawing this PR and abandoning https://github.com/deathcap/BedrockAPI

@basicer That was my reasoning as well. I didn't "strip the source", but rather wrote out stubs exclusively from the API documentation. Perhaps this is defensible, but unfortunately I'm too tired to pursue further, throwing in the towel now, but keeping the repo up in case you (or anyone else!) wants to pick up the torch and find out. Certainly I'd be interested in seeing what could be made of this, if it is crazy enough to work.

Oh and the code actually does compile, that's how far I got before submitting the PR (passes compile-test, dependency installed with `mvn install`, then compile with `./gradlew -x test`), just doesn't run, nor pass the tests. So to continue all that would need to be done is filling in the blanks, getting Glowstone (or Spigot?!) to a bootable state, writing a plugin loader, etc. Not for me, though. 

@SpaceManiac Thanks for the consideration, best regards

PR #187, opened 2014-09-07
Added: LightningStrikeEvent

anna-is-cute:
The most important event by far. If this should be done for the effect as well, let me know, and I'll update it.

### DIFF
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -11,6 +11,7 @@ import org.bukkit.*;
 import org.bukkit.block.Biome;
 import org.bukkit.block.Block;
 import org.bukkit.entity.*;
+import org.bukkit.event.weather.LightningStrikeEvent;
 import org.bukkit.event.weather.ThunderChangeEvent;
 import org.bukkit.event.weather.WeatherChangeEvent;
 import org.bukkit.generator.BlockPopulator;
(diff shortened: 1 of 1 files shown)


SpaceManiac:
I know it's a little confusing, but for very simple events (i.e. one-liners) like this one prefer creating event on-site & only using `EventFactory.callEvent`. Also I think it should be fired for effects too - plugins can differentiate with `event.getLightning().isEffect()`.

anna-is-cute:
Damn, that's how I had it, but I saw other events and changed it. Will fire for effects, too, then.

turt2live:
@jkcclemens this PR does not currently merge. Please fix that ASAP!

anna-is-cute:
Am currently grumbling about other PRs making my PR not able to merge. Will fix later tonight, hopefully.

anna-is-cute:
This can now be merged.

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1000,14 +1001,23 @@ public LivingEntity spawnCreature(Location loc, CreatureType type) {
         return (LivingEntity) spawn(loc, type.getEntityClass());
     }
 
+    private GlowLightningStrike strikeLightningFireEvent(final Location loc, final boolean effect) {
```
turt2live, line 1004: I don't think any of these variables need to be `final`, but that might just be me being too nitpicky.

turt2live:
Looks good otherwise. Passing this off to SpaceManiac for review/pull.

src/main/java/net/glowstone/GlowWorld.java
anna-is-cute, line 1004: Just my code style. Not harming anything, is it?

JeromSar, line 1004: @jkcclemens This should probably be in `EventFactory` named `onLightningStrike`

anna-is-cute, line 1004: No, it shouldn't. Look at the PR conversation. I had it like that originally.

JeromSar:
Related parent: #28

src/main/java/net/glowstone/GlowWorld.java
JeromSar, line 1004: @jkcclemens I see. Quite odd to events being called in many different places. That's probably not of the best interest to code clarity. Additionally, a lot of one-liners in called in `EventFactory` too. @SpaceManiac ?

SpaceManiac:
Merged in f78a5ed416672cb4c1a683e2f22622a9978bc9e3.

PR #189, opened 2014-09-07
Switch to Netty ReadTimeoutHandler

coelho:
The current method we use to timeout clients is based on the server tickrate. We should not use that, and instead use something built-in to Netty to stay reliable.

This is a work in progress, as we should also implement a new method of accomplishing the ping-pong process that is asynchronous. My reasoning behind this is that even if the server stops ticking to do a large amount of logic, your client should not disconnect due to a read timeout like you do in vanilla.

### DIFF
--- a/src/main/java/net/glowstone/net/GlowSession.java
+++ b/src/main/java/net/glowstone/net/GlowSession.java
@@ -43,12 +43,6 @@ import java.util.logging.Level;
 public final class GlowSession extends BasicSession {
 
     /**
-     * The number of ticks which are elapsed before a client is disconnected due
-     * to a timeout.
-     */
-    private static final int TIMEOUT_TICKS = 300;
-
-    /**
      * The server this session belongs to.
      */
     private final GlowServer server;
@@ -89,22 +83,11 @@ public final class GlowSession extends BasicSession {
     private String hostname;
 
     /**
-     * A timeout counter. This is increment once every tick and if it goes above
-     * a certain value the session is disconnected.
-     */
-    private int readTimeoutCounter = 0;
-
-    /**
      * Data regarding a user who has connected through a proxy, used to
      * provide online-mode UUID and properties and other data even if the
      * server is running in offline mode. Null for non-proxied sessions.
      */
     private ProxyData proxyData;
-
-    /**
-     * Similar to readTimeoutCounter but for writes.
-     */
-    private int writeTimeoutCounter = 0;
 
     /**
      * The player associated with this session (if there is one).
@@ -207,12 +190,26 @@ public final class GlowSession extends BasicSession {
     }
 
     /**
+     * Notify that the session is currently idle.
+     */
+    public void idle() {
+        if (pingMessageId == 0 && getProtocol() instanceof PlayProtocol) {
+            pingMessageId = random.nextInt();
+            if (pingMessageId == 0) {
+                pingMessageId++;
+            }
+            send(new PingMessage(pingMessageId));
+        } else {
+            disconnect("Timed out");
+        }
+    }
+
+    /**
      * Note that the client has responded to a keep-alive.
      * @param pingId The pingId to check for validity.
      */
     public void pong(long pingId) {
         if (pingId == pingMessageId) {
-            readTimeoutCounter = 0;
             pingMessageId = 0;
         }
     }
@@ -320,7 +317,6 @@ public final class GlowSession extends BasicSession {
 
     @Override
     public ChannelFuture sendWithFuture(Message message) {
-        writeTimeoutCounter = 0;
         if (!isActive()) {
             // discard messages sent if we're closed, since this happens a lot
             return null;
@@ -389,9 +385,6 @@ public final class GlowSession extends BasicSession {
      * Pulse this session, performing any updates needed.
      */
     void pulse() {
-        readTimeoutCounter++;
-        writeTimeoutCounter++;
-
         // drop the previous placement if needed
         if (previousPlacementTicks > 0 && --previousPlacementTicks == 0) {
             previousPlacement = null;
@@ -406,24 +399,6 @@ public final class GlowSession extends BasicSession {
             }
 
             super.messageReceived(message);
-            readTimeoutCounter = 0;
-        }
-
-        // let us know if the client has timed out yet
-        if (readTimeoutCounter >= TIMEOUT_TICKS) {
-            if (pingMessageId == 0 && getProtocol() instanceof PlayProtocol) {
-                pingMessageId = random.nextInt();
-                send(new PingMessage(pingMessageId));
-            } else {
-                disconnect("Timed out");
-            }
-            readTimeoutCounter = 0;
-        }
-
-        // let the client know we haven't timed out yet
-        if (writeTimeoutCounter >= TIMEOUT_TICKS && getProtocol() instanceof PlayProtocol) {
-            pingMessageId = random.nextInt();
-            send(new PingMessage(pingMessageId));
         }
     }
 
--- a/src/main/java/net/glowstone/net/pipeline/GlowChannelInitializer.java
+++ b/src/main/java/net/glowstone/net/pipeline/GlowChannelInitializer.java
@@ -1,14 +1,30 @@
 package net.glowstone.net.pipeline;
 
 import com.flowpowered.networking.ConnectionManager;
+
 import io.netty.channel.ChannelInitializer;
 import io.netty.channel.socket.SocketChannel;
+import io.netty.handler.timeout.IdleStateHandler;
+import io.netty.handler.timeout.ReadTimeoutHandler;
 import net.glowstone.net.protocol.ProtocolType;
 
 /**
  * Used to initialize the channels.
  */
 public final class GlowChannelInitializer extends ChannelInitializer<SocketChannel> {
+
+    /**
+     * The time in seconds which are elapsed before a client is disconnected due 
+     * to a read timeout.
+     */
+    private static final int READ_TIMEOUT = 20;
+
+    /**
+     * The time in seconds which are elapsed before a client is deemed idle due 
+     * to a write timeout.
+     */
+    private static final int WRITE_IDLE_TIMEOUT = 15;
+
     private final ConnectionManager connectionManager;
 
     public GlowChannelInitializer(ConnectionManager connectionManager) {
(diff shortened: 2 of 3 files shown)


src/main/java/net/glowstone/net/GlowSession.java
```diff
@@ -44,21 +47,15 @@
...
      */
     private final GlowServer server;
 
     /**
      * The Random for this session
-     */
+    */
     private final Random random = new Random();
-
+
```
turt2live: Whitespace

src/main/java/net/glowstone/net/GlowSession.java
```diff
@@ -104,7 +90,7 @@
      * The ID of the last ping message sent, used to ensure the client responded correctly.
      */
     private int pingMessageId;
-
+
```
turt2live: Whitespace

src/main/java/net/glowstone/net/GlowSession.java
```diff
@@ -170,16 +156,30 @@ public void setVerifyUsername(String verifyUsername) {
...
      * @param pingId The pingId to check for validity.
      */
     public void pong(long pingId) {
         if (pingId == pingMessageId) {
-            readTimeoutCounter = 0;
             pingMessageId = 0;
         }
     }
-
+
```
turt2live: Whitespace

src/main/java/net/glowstone/net/pipeline/GlowChannelInitializer.java
```diff
@@ -1,14 +1,30 @@
...
 import io.netty.channel.socket.SocketChannel;
+import io.netty.handler.timeout.IdleStateHandler;
+import io.netty.handler.timeout.ReadTimeoutHandler;
 import net.glowstone.net.protocol.ProtocolType;
 
 /**
  * Used to initialize the channels.
  */
 public final class GlowChannelInitializer extends ChannelInitializer<SocketChannel> {
+
```
turt2live: Whitespace

src/main/java/net/glowstone/net/pipeline/GlowChannelInitializer.java
```diff
@@ -1,14 +1,30 @@
...
  * Used to initialize the channels.
  */
 public final class GlowChannelInitializer extends ChannelInitializer<SocketChannel> {
+    
+    /**
+     * The time in seconds which are elapsed before a client is disconnected due 
+     * to a read timeout.
+     */
+    private static final int READ_TIMEOUT = 20;
+
```
turt2live: Whitespace

src/main/java/net/glowstone/net/pipeline/GlowChannelInitializer.java
```diff
@@ -1,14 +1,30 @@
...
+     * to a read timeout.
+     */
+    private static final int READ_TIMEOUT = 20;
+    
+    /**
+     * The time in seconds which are elapsed before a client is deemed idle due 
+     * to a write timeout.
+     */
+    private static final int WRITE_IDLE_TIMEOUT = 15;
+
```
turt2live: Whitespace

src/main/java/net/glowstone/net/pipeline/MessageHandler.java
```diff
@@ -48,6 +52,13 @@ public void channelInactive(ChannelHandlerContext ctx) {
     protected void channelRead0(ChannelHandlerContext ctx, Message i) {
         session.get().messageReceived(i);
     }
+
```
turt2live: Whitespace

src/main/java/net/glowstone/net/pipeline/MessageHandler.java
```diff
@@ -48,6 +52,13 @@ public void channelInactive(ChannelHandlerContext ctx) {
     protected void channelRead0(ChannelHandlerContext ctx, Message i) {
         session.get().messageReceived(i);
     }
+    
+    @Override
+    public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
+        if (evt instanceof IdleStateEvent) {
+            ((GlowSession) session.get()).idle(); // todo: there needs to be a better way
```
turt2live: Ideas? (re: todo)

src/main/java/net/glowstone/net/GlowSession.java
```diff
@@ -170,16 +156,30 @@ public void setVerifyUsername(String verifyUsername) {
...
+            pingMessageId = random.nextInt();
+            if (pingMessageId == 0) {
+                pingMessageId++;
+            }
+            send(new PingMessage(pingMessageId));
+        } else {
+            disconnect("Timed out");
+        }
+    }
+
```
turt2live: Whitespace

src/main/java/net/glowstone/net/GlowSession.java
```diff
@@ -44,21 +47,15 @@
...
-
-    /**
      * The server this session belongs to.
      */
     private final GlowServer server;
 
     /**
      * The Random for this session
-     */
+    */
```
turt2live: Incorrect alignment

EntityReborn:
This is still marked WIP, what is the status of this?

coelho:
There's still something in flow-net that makes the code behind this PR not entirely uniform. Waiting to see if I can get a solution to that first.

turt2live:
Sorry - bot fail :(

turt2live:
@coelho This is not mergable and still marked as a WIP. Is there something that needs to be done or can this be reviewed fully? 

The mergability problem will need to be fixed before review can continue. Thanks!

ghost:
Offtopic: Why minecraft need alive packet then? :D

coelho:
@ensirius The keep alive packet is used when no data is being sent - IE, we detect that the connection is about to read timeout, so we send the packet.

On a different note, other than the merge conflicts which I'll fix in a bit, and the fact we do an odd cast in GlowSession, this is a pretty solid implementation.

ghost:
@coelho time update packet sends every second

coelho:
@ensirius If the server is lagging due to a large amount of processing on the world thread, then no, the client will not receive a time update packet every second.

This is why in vanilla Minecraft you get disconnected for read timeouts when the world thread has a large amount of work.

turt2live:
@coelho This is looking pretty inactive code-wise. If after the merge conflicts are resolved this PR is done, please remove "WIP" from the title to have it reviewed. 

If this is still a work in progress, please comment back stating so. 

Thanks!

turt2live:
@coelho Please respond to my last message. Thanks!

coelho:
@turt2live

Should be up to date, and will also fix #411 to my knowledge.

turt2live:
Going to look at this when I get the chance :) Thanks.

dequis:
I was getting timeouts when doing save-all from console, merging this PR fixed it.

Additionally, I lost 20kg of weight, stopped losing hair, managed to pay off my student loans and got a girlfriend who loves me even though she's not real.

So yeah this PR is pretty good.

dequis:
_"What can I say about the Netty ReadTimeoutHandler that hasn't already been said about the wheel, penicillin, or the iPhone? This is one of the greatest inventions of all time."_
-- Mrs T.

---

_"I believe that testing this PR has made me a better man, which is remarkable because, well....I'm a chick.*_
-- Diane P.

---

_"The Netty ReadTimeoutHandler power is obvious. This is living proof that you will get women, and fly. Most importantly my son was born without bones and when I tried this timeout handler with him he grew bones. Don't ask me how it happened but the magic is there. I wish I could hug the author of this PR and thank them for everything they have done for my family"_
 -- Dustin D.

---

_"So I got this ReadTimeoutHandler because of, you know, the sweet timeout handler in it. However, having tested this ReadTimeoutHandler for three weeks now and having tried it out in a variety of situations, both formal and informal, I'm beginning to believe that some of the benefits ---- as described by other reviewers ---- are exaggerated. For example, not ONE supermodel has approached me. Some of you may be used to having supermodels approach you on a regular basis but, believe me, I am not: I would notice one should she appear in my vicinity."_
-- Go Down M.

---

_"I must admit, that when I first tested this PR, I thought it was the very pinnacle of human achievement. And I was bursting with pride that this great nation created such a marvel. But then, on a missionary trip to Africa, I saw kids so poor, they had no timeout handlers on their servers. No handlers at all. Yet I had a Netty one. I wept in shame._

_Upon reflection, I wondered: What must the lesser nations of the world (like Ethiopia, Uganda, France and any of the crappy ones that end in '-slovakia' or '-istan') think of us when they go to their beds at night having no timeout handlers on their servers, while we have the Netty one? Our arrogance must now sting their pride even greater than the long silent whips of their colonial masters. I could sense an underlying envy that hinted at a hidden, yet simmering rage. How could we keep such wealth to ourselves, while others suffered? How could we not share this wonderful innovation with the world?"_

-- John F. K.

---

<sub><sub>seriously though merge this please</sub></sub>

mastercoms:
Yes, merge this.

Emiel45:
:+1:

SpaceManiac:
From the documentation, it seems like `IdleStateHandler` can perform dual duty for read and write timeouts. Is there a specific reason you chose to have a separate `ReadTimeoutHandler`?

coelho:
IdleStateHandler is only for informing the pipeline that a connection has become idle. This functionality is only needed for writes as we use it to send keepalives (to keep the connection open even though we're not sending anything)

As for read timeouts, whenever a timeout hits we basically always close the connection. There's no special behavior that is needed and if there is, something in the pipeline can just catch the exception knowing that the connection is closed. On top of this, IdleStateHandler allows us to disable the read idle portion completely making this really not even affect performance much (if at all).

https://github.com/netty/netty/blob/master/handler/src/main/java/io/netty/handler/timeout/IdleStateHandler.java#L44

We can always use the IdleStateHandler to replicate the functionality that the ReadTimeoutHandler already provides to us, but I'm not sure if there's a need.

SpaceManiac:
Alright, on further review I better understand exactly what's happening here. However, I still need to consider whether this is the desired behavior. Old behavior will ping the client on read timeout, giving them the opportunity to respond and break the timeout if the client is merely quiet rather than dead (though no compliant client should do this - so perhaps your behavior is better). The write timeout handling can probably be simplified in that case to simply send the ping messages with less extra handling than it currently has, since the old idle logic is to handle read timeouts and all the write timeout is intended to do is prevent the client from reaching a read timeout.

coelho:
This is mostly an improvement over Vanilla, where in Vanilla if there is a large amount of lag on the main thread, clients will simply disconnect from the server. With this PR, clients will stay connected throughout the lag spike.
Although it's not Vanilla behavior, it is definitely more effective than Vanilla.

turt2live:

PR #190, opened 2014-09-07
Handle status packets asynchronously

coelho:
Status packets are currently behind handled on the main thread. With this PR, they are moved off of the main thread and are handled by the NIO threads.

It is good to note however that ServerListPingEvent, although incorrectly documented, is an asynchronously fired event within currently available Bukkit implementations.

### DIFF
--- a/src/main/java/net/glowstone/net/message/status/StatusRequestMessage.java
+++ b/src/main/java/net/glowstone/net/message/status/StatusRequestMessage.java
@@ -8,5 +8,5 @@ public final class StatusRequestMessage implements AsyncableMessage {
     public boolean isAsync() {
         return true;
     }
-    
+
 }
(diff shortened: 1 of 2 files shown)


src/main/java/net/glowstone/net/message/status/StatusRequestMessage.java
```diff
@@ -1,6 +1,12 @@
...
+import com.flowpowered.networking.AsyncableMessage;
 
-public final class StatusRequestMessage implements Message {
+public final class StatusRequestMessage implements AsyncableMessage {
+
+    @Override
+    public boolean isAsync() {
+        return true;
+    }
+
```
turt2live: Whitespace

turt2live:
Sorry - bot fail :(

turt2live:
Looks good!

SpaceManiac:
Merged in 0c6e09dd4a087b32f803e2ab14457a2a18daf82b.

Issue #172, opened 2014-09-06
PlayerSwingArmsMessage


Kaskadee:
Sometimes when trying to place a block an error occurs:

``` java
Error while handling PlayerSwingArmMessage (handler: PlayerSwingArmHandler)
java.lang.IllegalStateException: Start block missed in BlockIterator
        at org.bukkit.util.BlockIterator.<init>(BlockIterator.java:170)
        at org.bukkit.util.BlockIterator.<init>(BlockIterator.java:233)
        at org.bukkit.util.BlockIterator.<init>(BlockIterator.java:268)
        at net.glowstone.entity.GlowLivingEntity.getLineOfSight(GlowLivingEntity
.java:205)
        at net.glowstone.entity.GlowLivingEntity.getTargetBlock(GlowLivingEntity
.java:231)
        at net.glowstone.net.handler.play.player.PlayerSwingArmHandler.handle(Pl
ayerSwingArmHandler.java:17)
        at net.glowstone.net.handler.play.player.PlayerSwingArmHandler.handle(Pl
ayerSwingArmHandler.java:13)
        at com.flowpowered.networking.session.BasicSession.handleMessage(BasicSe
ssion.java:80)
        at com.flowpowered.networking.session.BasicSession.messageReceived(Basic
Session.java:139)
        at net.glowstone.net.GlowSession.pulse(GlowSession.java:363)
        at net.glowstone.net.SessionRegistry.pulse(SessionRegistry.java:23)
        at net.glowstone.scheduler.GlowScheduler.pulse(GlowScheduler.java:172)
        at net.glowstone.scheduler.GlowScheduler.access$100(GlowScheduler.java:2
4)
        at net.glowstone.scheduler.GlowScheduler$2.run(GlowScheduler.java:109)
        at java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
        at java.util.concurrent.FutureTask.runAndReset(Unknown Source)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.
access$301(Unknown Source)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.
run(Unknown Source)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
        at java.lang.Thread.run(Unknown Source)
```

PR #192, opened 2014-09-07
Handle target blocks gracefully.

JeromSar:
Sometimes, player.getTargetBlock() returns an IllegalStateException.
This is usually the case if players stand at height 127 and look up when left-clicking.
The exception can be safely caught and treated as an invalid block (air in this case).

### DIFF
--- a/src/main/java/net/glowstone/net/handler/play/player/PlayerSwingArmHandler.java
+++ b/src/main/java/net/glowstone/net/handler/play/player/PlayerSwingArmHandler.java
@@ -14,7 +14,13 @@ public final class PlayerSwingArmHandler implements MessageHandler<GlowSession, 
     public void handle(GlowSession session, PlayerSwingArmMessage message) {
         final GlowPlayer player = session.getPlayer();
 
-        Block block = player.getTargetBlock(null, 6);
+        Block block;
+        try {
+            block = player.getTargetBlock(null, 6);
+        } catch (IllegalStateException ex) {
+            block = null;
+        }
+
         if (block == null || block.isEmpty()) {
             if (EventFactory.onPlayerInteract(player, Action.LEFT_CLICK_AIR).isCancelled())
                 return;


turt2live:
shouldn't the root cause be identified?

JeromSar:
@turt2live The root cause is in fact the Bukkit API itself (see https://github.com/Bukkit/Bukkit/blob/master/src/main/java/org/bukkit/util/BlockIterator.java#L170). This exception is thrown when the first block of the iteration can not be found. For example, when a player looks up, the first block would technically be outside the modifiable world.

For this particular case, I don't think it's necessary to change the Bukkit API. The change this pull request is minimal and avoids potential (and probable) errors when players swing their arms.

turt2live:
It should be fixed in the Bukkit API for plugins calling the method...

JeromSar:
@turt2live That would be an API breaking change. Additionally, it would require plugins to use the GlowStone-specific Bukkit, which is not GlowStone's goal (to be compatible with Bukkit plugins).

turt2live:
@JeromSar This is breaking?

```
// Plugin code
try{
    Block block = null; // Return value from method in Bukkit
    block.stuff(); // throws NPE
}catch(Exception e){
    // ;o here we are
}
```

JeromSar:
@turt2live You were a suggesting we fix the error in Bukkit, correct?

turt2live:
Correct.

JeromSar:
@turt2live Some plugins might depend on that error being thrown.

In this case, the `IllegalStateException` is purposely thrown. Catching the exception in Bukkit would be an API breaking change. 

See: https://github.com/Bukkit/Bukkit/blob/master/src/main/java/org/bukkit/util/BlockIterator.java#L170

turt2live:
I don't think a try/catch in Glowstone is suitable. Instead, Bukkit should be resolved or better handling on Glowstone's end should be implemented.

Currently in Bukkit there is no documentation on that exception being thrown.

SpaceManiac:
Would love to see some concrete evidence that both plugins which don't handle this case themselves and plugins which do are likely to benefit from this change going into a Bukkit fork.

JeromSar:
@SpaceManiac I don't believe this change should be implemented in Bukkit for that reason. For backwards compatibility, it's best to avoid touching Bukkit whenever possible.

@turt2live I could implement better pre-checking on GlowStone's end but that would imply a performance loss. It would be inefficient too as it would check if the starting block exists twice. This change is both simpler and future- and backwards compatible.

turt2live:
It doesn't change the fact that it's a workaround. Workarounds are not ideal if a better solution can be implemented (and if there is no better solution, the problem should be looked at in more detail).

turt2live:
Ticket: #172

turt2live:
Sorry - bot fail :(

turt2live:
After much deliberation with myself, I can't think of a better solution that would be less... meeeehhhh.

Looks good, I suppose.

SpaceManiac:
Merged in 367f1b861861128a25fbe2d1ef44f1142822764e.

PR #193, opened 2014-09-07
Let the player drop the (whole) ItemStack in his hand with (Ctrl+)Q

Tonodus:
When you press (Ctrl+)Q in minecraft, you are able to drop the (whole) ItemStack, just like if you drag and drop it out of your inventory.

### DIFF
--- a/src/main/java/net/glowstone/EventFactory.java
+++ b/src/main/java/net/glowstone/EventFactory.java
@@ -7,7 +7,6 @@ import org.bukkit.ChatColor;
 import org.bukkit.Location;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
-import org.bukkit.entity.Item;
 import org.bukkit.entity.Player;
 import org.bukkit.event.Event;
 import org.bukkit.event.block.Action;
@@ -67,10 +66,6 @@ public final class EventFactory {
     ////////////////////////////////////////////////////////////////////////////
     // Player Events
 
-    public static PlayerDropItemEvent onPlayerDropItem(Player player, Item droppedItem) {
-        return callEvent(new PlayerDropItemEvent(player, droppedItem));
-    }
-    
     public static AsyncPlayerPreLoginEvent onPlayerPreLogin(String name, InetSocketAddress address, UUID uuid) {
         // call async event
         final AsyncPlayerPreLoginEvent event = new AsyncPlayerPreLoginEvent(name, address.getAddress(), uuid);
--- a/src/main/java/net/glowstone/EventFactory.java
+++ b/src/main/java/net/glowstone/EventFactory.java
@@ -7,6 +7,7 @@ import org.bukkit.ChatColor;
 import org.bukkit.Location;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
+import org.bukkit.entity.Item;
 import org.bukkit.entity.Player;
 import org.bukkit.event.Event;
 import org.bukkit.event.block.Action;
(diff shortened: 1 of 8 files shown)


SonarBeserk:
It works but the tossing strength is weaker than it should be. The lack of an arc to item drops is very noticeable as well.

SonarBeserk:
This won't be useful to merge until item pickup is implemented.

Tonodus:
Well, I agree with you, that the visuality wasn't that nice. I changed it to look closer to minecraft.
However, I don't think it is useless to implement dropping items, even if it is impossible to pick them up. Tell me, why it is more useful to implement picking up items -even if you can't drop them- first.

SonarBeserk:
Should have said as useful sorry, it is definitely good but pick up is also still needed.

turt2live:
Sorry - bot fail :(

turt2live:
@Tonodus This has a bug where drops "lag" and repeat animation. What happens is that you go around throwing items, but after a while they "lurch" back to their thrown origin point and fall straight down. One easy way to demonstrate this is to go into creative mode with a stack of dirt and fly in a circle spamming Q. Land on the ground facing the circle and wait until they "lurch" and fall straight down.

Tonodus:
@turt2live No that's not a bug in my code, but in GlowEntity.
If you set an entity's velocity, it will only move on the client, but the location remains on the server. The server resends the current position of each entity after some time and that's why you se the item teleported to the origin location.
To fix this bug, there must be gravity and velocity support server side. But if this is done, my code will automatically work as expect.

The PR fixing the mentioned problem: #196

Tonodus:
It still says that this PR needs changes. I'd like to know about them, or a statement that the work for this PR is finished on my side.

turt2live:
Considering this has been sitting for a bit, some testing has been done. Here are the bugs:
- (minor) Dropping an item stack with `Q` in survival mode makes the item in hand "flicker" (clear slot replacement)
- Players cannot click (left or right) outside the window to drop items in creative mode
- When holding an item and using `escape` to close the inventory, items are not dropped in creative or adventure mode
  - Creative mode swallows the items, as expected

I'm not sure what @SpaceManiac wants out of this PR, but those are the cases that do not match vanilla with this PR.

Tonodus:
@turt2live @SpaceManiac 
I implemented the things turt2live mentioned above, expect the first point (I don't understand what the issue is).

turt2live:
When tested, simply pressing Q while in survival mode would make the item "flicker" in-hand. There's no real way to explain this beyond just trying it yourself.

Tonodus:
@turt2live just to clarify: are you sure, this isn't some sort of client side bug (happening also in vanilla)? Reducing the amount of an item in hand, let's the client 'switch' the item to itself, as if you changed it to a new one with a different type, but that also appears in vanilla. Not sure, whether you thought of this.

dequis:
Simple imports conflicts in GlowHumanEntity and GlowPlayer, just keep in mind the renamed net.glowstone.entity.meta.profile.PlayerProfile. [Branch](https://github.com/dequis/Glowstone/compare/pr-193)

turt2live:
Changes don't appear to have changed PR content. Considering @SpaceManiac keeps taking this from me - any updates?

Tonodus:
@turt2live No, only rebasing, because I'm still think the glitch happens in vanilla too and is not fixable.

dequis:
Can confirm that the item "switches" to itself in both vanilla and glowstone. So, looks good to me.

Any comments on why this is "needs further review" and assigned to @SpaceManiac would be appreciated.

Also I think we've used "ready to pull" + "needs further review" in the past, could be used here I guess

turt2live:
@dequis it's as "needs further review" so that it stays assigned to @SpaceManiac as he kept removing me as an assignee in the past.

dequis:
But other than that, it looks good to you, right?

turt2live:
I have not done my usual in-depth review.

SpaceManiac:
Merged in 3997385b73133601f67e1dfabe687b6ccbcda372. Thanks!

PR #194, opened 2014-09-07
Chest animations (Open/Close)

wow4201:
Sends the chest open and close animations. Tested, works as intended. Player can only open an unopened chest. Chest only closes when the player leaves the chest inventory.

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockChest.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockChest.java
@@ -2,17 +2,22 @@ package net.glowstone.block.blocktype;
 
 import net.glowstone.GlowChunk;
 import net.glowstone.GlowServer;
+import net.glowstone.block.GlowBlock;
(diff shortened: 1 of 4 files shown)


Ribesg:
Indentation is broken. Also you should cast only once, store a local reference to the cast if multiple uses are needed.

nevercast:
Where are the players tracked when multiple players have a chest open?

If Alice opens the chest and then Bob starts using it also, then Alice closes it, the chest should not animate closed until Bob has finished.

dequis:
This is a dead ticket, you know. wow4201 already ragequit weeks ago because he didn't feel like updating his stuff and blamed that on our PR handling.

nevercast:
If the PR comes up again this one will likely be referenced, at which point they should take note that it needs to handle multiple players interacting with it at the same time. Earlier MC versions used a counter, later ones I believe track player GUIDs.

Though it was a bit of a grave dig, I'll keep a closer eye on dates.

PR #197, opened 2014-09-07
Chest animation

wow4201:
Player Chest/EnderChest Animations (Open/Close) to nearby players, Works as intended, can only open if chest is closed, can only close when player closes out of storage inventory.

### DIFF
--- a/src/main/java/net/glowstone/block/state/GlowEnderChest.java
+++ b/src/main/java/net/glowstone/block/state/GlowEnderChest.java
@@ -1,0 +1,63 @@
+package net.glowstone.block.state;
+
+import net.glowstone.GlowChunk;
+import net.glowstone.block.GlowBlock;
+import net.glowstone.block.GlowBlockState;
+import net.glowstone.block.entity.TEEnderChest;
+import net.glowstone.entity.GlowPlayer;
+
+import org.bukkit.Location;
+import org.bukkit.Material;
+import org.bukkit.block.Chest;
+import org.bukkit.inventory.Inventory;
+
+public class GlowEnderChest extends GlowBlockState implements Chest {
+
+	public int state;
+    public GlowEnderChest(GlowBlock block) {
+        super(block);
+        state = 0;
+    }
+
+    private TEEnderChest getTileEntity() {
+        return (TEEnderChest) getBlock().getTileEntity();
+    }
+
+    @Override
+    public Inventory getBlockInventory() {
+        return getTileEntity().getInventory();
+    }
+
+    @Override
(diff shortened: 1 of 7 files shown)


wow4201:
I'll clean the formatting up tommorow.

TomCaserta:
Had a quick look through either Im understanding it incorrectly or does this only allow one player in a chest at a time?

src/main/java/net/glowstone/block/state/GlowEnderChest.java
```diff
@@ -0,0 +1,63 @@
...
+import net.glowstone.block.GlowBlockState;
+import net.glowstone.block.entity.TEEnderChest;
+import net.glowstone.entity.GlowPlayer;
+
+import org.bukkit.Location;
+import org.bukkit.Material;
+import org.bukkit.block.Chest;
+import org.bukkit.inventory.Inventory;
+
+public class GlowEnderChest extends GlowBlockState implements Chest {
```
EntityReborn, line 14: Could one not extend GlowChest? Lots of code duplication here.

EntityReborn:
@TomCaserta this does not seem to be the case to me.

src/main/java/net/glowstone/block/state/GlowEnderChest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public Inventory getBlockInventory() {
+        return getTileEntity().getInventory();
+    }
+
+    @Override
+    public Inventory getInventory() {
+        // todo: handle double chests
+        return getBlockInventory();
```
TomCaserta, line 34: Shouldnt this be returning player.getEnderChest() ?

EntityReborn, line 34: That would probably be correct. This would depend on the implementation of internals. It might be better to modify `TEEnderChest`'s `getInventory()` method.

TomCaserta:
Ah sorry just tested it and you are correct. I must be reading the code wrong because I thought it was setting a flag when the chests open and only opening if that flags not set but yeah.

Also it appears that if someone else opens the same chest whilst another is viewing it will not stay open if the someone closes the inventory view. You could use getViewers() on the inventory to ensure that there is no one else viewing the inventory before sending the chest close animation.

turt2live:
~~Double chests don't seem to work.~~

nvm. I'm dumb

EntityReborn:
Why was this closed?

Johni0702:
This was put into one pull request with far less commits: #211

PR #198, opened 2014-09-07
Added: Buttons

JeromSar:
Added buttons. They are placed and updated properly.

Tested and well. :)

### DIFF
--- a/src/main/java/net/glowstone/block/ItemTable.java
+++ b/src/main/java/net/glowstone/block/ItemTable.java
@@ -85,8 +85,6 @@ public final class ItemTable {
         reg(Material.WEB, new BlockDirectDrops(Material.STRING));
         reg(Material.FIRE, new BlockFire());
         reg(Material.MONSTER_EGGS, new BlockDropless());
-        reg(Material.STONE_BUTTON, new BlockButton(Material.STONE_BUTTON));
-        reg(Material.WOOD_BUTTON, new BlockButton(Material.WOOD_BUTTON));
         reg(Material.FURNACE, new BlockFurnace());
         reg(Material.LEVER, new BlockLever());
         reg(Material.HOPPER, new BlockHopper());
--- a/src/main/java/net/glowstone/block/ItemTable.java
+++ b/src/main/java/net/glowstone/block/ItemTable.java
@@ -84,6 +84,8 @@ public final class ItemTable {
         reg(Material.WEB, new BlockDirectDrops(Material.STRING));
         reg(Material.FIRE, new BlockDropless());
         reg(Material.MONSTER_EGGS, new BlockDropless());
+        reg(Material.STONE_BUTTON, new BlockButton(Material.STONE_BUTTON));
+        reg(Material.WOOD_BUTTON, new BlockButton(Material.WOOD_BUTTON));
 
         reg(Material.SIGN, new ItemSign());
         reg(Material.REDSTONE, new ItemPlaceAs(Material.REDSTONE_WIRE));
--- a/src/main/java/net/glowstone/block/ItemTable.java
+++ b/src/main/java/net/glowstone/block/ItemTable.java
@@ -106,6 +106,7 @@ public final class ItemTable {
         reg(Material.QUARTZ_BLOCK, new BlockQuartz());
         reg(Material.LOG, new BlockLog());
         reg(Material.LOG_2, new BlockLog2());
+
         reg(Material.SIGN, new ItemSign());
         reg(Material.REDSTONE, new ItemPlaceAs(Material.REDSTONE_WIRE));
         reg(Material.SUGAR_CANE, new ItemPlaceAs(Material.SUGAR_CANE_BLOCK));
--- a/src/main/java/net/glowstone/block/blocktype/BlockAttachable.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockAttachable.java
@@ -15,27 +15,27 @@ public class BlockAttachable extends BlockType {
         byte data = state.getRawData();
         switch (attachedFace) {
             case UP: {
-                data = 0;
+                data |= 0;
                 break;
             }
             case WEST: {
-                data = 1;
+                data |= 1;
                 break;
             }
             case EAST: {
-                data = 2;
+                data |= 2;
                 break;
             }
             case NORTH: {
-                data = 3;
+                data |= 3;
                 break;
             }
             case SOUTH: {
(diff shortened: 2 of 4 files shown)


src/main/java/net/glowstone/block/blocktype/BlockButton.java
```diff
@@ -0,0 +1,91 @@
...
+
+public class BlockButton extends BlockType {
+
+    public BlockButton(Material material) {
+        setDrops(new ItemStack(material));
+    }
+
+    private void setAttachedFace(Button button, BlockFace attachedFace) {
+        byte data = button.getData();
+        switch (attachedFace) {
```
turt2live: Not sure on the coding standards of the project, but this might look better as single lines.

```
switch(attachedFace){
    case UP: data |= 0; break;
    case WEST: data |= 1; break;
    // ...
}
```

src/main/java/net/glowstone/block/blocktype/BlockButton.java
```diff
@@ -0,0 +1,91 @@
...
+    public void placeBlock(GlowPlayer player, GlowBlockState state, BlockFace face, ItemStack holding, Vector clickedLoc) {
+        super.placeBlock(player, state, face, holding, clickedLoc);
+
+        final MaterialData data = state.getData();
+        if (!(data instanceof Button)) {
+            GlowServer.logger.warning("Placing " + getMaterial().name() + ": MaterialData was of wrong type (" + data.getClass().getName() + ")");
+            return;
+        }
+
+        final Button button = (Button) data;
```
turt2live: You only use this once and the whitespace makes it look somewhat weird. Maybe just rid yourself of the variable? Also it doesn't need to be final.

src/main/java/net/glowstone/block/blocktype/BlockButton.java
```diff
@@ -0,0 +1,91 @@
...
+import org.bukkit.scheduler.BukkitRunnable;
+import org.bukkit.util.Vector;
+
+public class BlockButton extends BlockType {
+
+    public BlockButton(Material material) {
+        setDrops(new ItemStack(material));
+    }
+
+    private void setAttachedFace(Button button, BlockFace attachedFace) {
```
turt2live: This method may be better used in an `Attachable` interface or similar. Pretty sure everything in MC uses the same data values for now.

src/main/java/net/glowstone/block/blocktype/BlockButton.java
instantdelay: Not that my opinion matters but `final` should really be the default. Unless there's a reason for taking final away, immutability is usually better.

turt2live: Beyond the convention, nope.

src/main/java/net/glowstone/block/blocktype/BlockButton.java
JeromSar: @turt2live That would imply changing the other similar implementations too. Might be something for a future pull request. :)

src/main/java/net/glowstone/block/blocktype/BlockButton.java
JeromSar: @turt2live You're right. Fixed.

src/main/java/net/glowstone/block/blocktype/BlockButton.java
turt2live: I would recommend doing it now so we don't need to review another PR

src/main/java/net/glowstone/block/blocktype/BlockButton.java
turt2live: I lied. Oracle conventions say otherwise. Sorry :(

JeromSar: @turt2live Alright, will fix again. :)

zD12:
It appears to work.

src/main/java/net/glowstone/block/ItemTable.java
```diff
@@ -107,7 +106,6 @@ private void registerBuiltins() {
         reg(Material.QUARTZ_BLOCK, new BlockQuartz());
         reg(Material.LOG, new BlockLog());
         reg(Material.LOG_2, new BlockLog2());
-
```
turt2live, line 110: This was here for a reason, please put it back.

src/main/java/net/glowstone/block/blocktype/BlockAttachable.java
```diff
@@ -0,0 +1,45 @@
...
+import static org.bukkit.block.BlockFace.WEST;
+import org.bukkit.block.BlockState;
+
+public class BlockAttachable extends BlockType {
+
+    public void setAttachedFace(BlockState state, BlockFace attachedFace) {
+        byte data = state.getRawData();
+        switch (attachedFace) {
+            case UP: {
+                data |= 0;
```
turt2live, line 18: These should not be `|=`, please use `=`

src/main/java/net/glowstone/block/ItemTable.java
JeromSar, line 110: @turt2live What's the reason?

turt2live, line 110: Separates blocks and items.

src/main/java/net/glowstone/block/blocktype/BlockAttachable.java
JeromSar, line 18: @turt2live I took example from @jkcclemens in https://github.com/GlowStoneMC/Glowstone/commit/8d9cf2febb247e0bbf011ca6f5faa92f1aaf244b#diff-53a5eb7d2e87a87b269421a3e89446a8R20

Seemed odd but I went with it.
Tweaked as requested now. :)

turt2live, line 18: I think I see why it was done like that... 

There's actually a bug I just accidentally asked you to add XD. Sorry ._.

JeromSar, line 18: @turt2live Haha, alright. I'll revert it.

turt2live:
Sorry - bot fail :(

turt2live:
Looks good!

SpaceManiac:
Merged in 32496b095cdb7f59984f781fd018d7458d009c98.

PR #200, opened 2014-09-07
Open blocks (fence gates, trap doors) with right click

Tonodus:
A new BlockType that supports opening the blocks if the player right clicks them.
It uses the Openable-MaterialData to open/close.

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockFenceGate.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockFenceGate.java
@@ -9,6 +9,7 @@ import org.bukkit.block.BlockState;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.material.Gate;
 import org.bukkit.material.MaterialData;
+import org.bukkit.material.Openable;
 import org.bukkit.util.Vector;
 
 public class BlockFenceGate extends BlockOpenable {
@@ -37,22 +38,13 @@ public class BlockFenceGate extends BlockOpenable {
     }
 
     @Override
-    public boolean blockInteract(GlowPlayer player, GlowBlock block, BlockFace face, Vector clickedLoc) {
-        boolean changed = super.blockInteract(player, block, face, clickedLoc);
-
-        if (changed) {
-            BlockState state = block.getState();
-            MaterialData data = state.getData();
-            if (data instanceof Gate) {
-                Gate gate = (Gate) data;
-                gate.setFacingDirection(getOpenDirection(player, gate.getFacing()));
-                state.update(true);
-            } else {
-                warnMaterialData(Gate.class, data);
-            }
+    protected void onOpened(GlowPlayer player, GlowBlock block, BlockFace face, Vector clickedLoc, GlowBlockState state, MaterialData materialData) {
+        if (materialData instanceof Gate) {
+            Gate gate = (Gate) materialData;
+            gate.setFacingDirection(getOpenDirection(player, gate.getFacing()));
+        } else {
+            warnMaterialData(Gate.class, materialData);
         }
-
-        return changed;
     }
 
     private static BlockFace getOpenDirection(GlowPlayer player, BlockFace oldFacing) {
--- a/src/main/java/net/glowstone/block/blocktype/BlockFenceGate.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockFenceGate.java
@@ -23,7 +23,7 @@ public class BlockFenceGate extends BlockOpenable {
                 gate.setFacingDirection(getOpenDirection(player, gate.getFacing()));
                 state.update(true);
             } else {
-                GlowServer.logger.warning("Called blockInteract on BlockFenceGate, however block is type of " + data);
+                warnMaterialData(Gate.class, data);
             }
         }
 
--- a/src/main/java/net/glowstone/block/blocktype/BlockFenceGate.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockFenceGate.java
@@ -2,14 +2,38 @@ package net.glowstone.block.blocktype;
 
 import net.glowstone.GlowServer;
 import net.glowstone.block.GlowBlock;
+import net.glowstone.block.GlowBlockState;
 import net.glowstone.entity.GlowPlayer;
 import org.bukkit.block.BlockFace;
 import org.bukkit.block.BlockState;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.material.Directional;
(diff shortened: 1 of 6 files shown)


EntityReborn:
As this is a relatively small update, might be worth it to squash your commits.

Tonodus:
Sorry, do you mean I should squash these four commits into one? I'm going to do that (next time?). I'm relatively new with git, didn't noticed that woul be better.

EntityReborn:
Yes, that is what I mean. Regardless, there isn't an established guideline for this for PRs, besides commits getting squashed right before integration. Up to you.

Tonodus:
Issue: Fence gates don't open themselves off from the player

Tonodus:
Fixed issue, but feel free to comment my solution.

src/main/java/net/glowstone/block/blocktype/BlockFenceGate.java
```diff
@@ -0,0 +1,49 @@
...
+import net.glowstone.block.GlowBlock;
+import net.glowstone.entity.GlowPlayer;
+import org.bukkit.block.BlockFace;
+import org.bukkit.block.BlockState;
+import org.bukkit.material.Gate;
+import org.bukkit.material.MaterialData;
+import org.bukkit.util.Vector;
+
+/**
+ * Created by Tonodus (http://tonodus.github.io) on 08.09.2014.
```
daboross: I don't _think_ we want to include things like this. Could you remove this line?

daboross:
Here's a guide to squashing, if you need one: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html.

src/main/java/net/glowstone/block/blocktype/BlockFenceGate.java
Tonodus: I apologize, that's just my IDE's default header I read over.

Tonodus:
Removed header and squashed commits.

EntityReborn:
This currently has merge conflicts. Please resolve.

Tonodus:
Done

turt2live:
Sorry - bot fail :(

turt2live:
@Tonodus this works for standard fence gates and trap doors, but not the extended gates. You still end up placing a block in front of the varied gates (birch, etc)

Tonodus:
@turt2live: Everything should be ok now, sorry for the inconveniences.

SpaceManiac:
Looks pretty good for the most part, just a couple of things with the fence gates:
- The orientation should only be changed when the gate is opened (otherwise there's an animation tic when closing a fence gate from some directions).
- I _think_ the gate shouldn't be reoriented for certain player angles (near perpendicular to the gate's axis) and its old orientation should just be used, at least from what I can see from opening it from this perspective, but I'm not certain.

Tonodus:
@SpaceManiac I fixed all you mentioned:
- I couldn't reproduce your animation tic, but the orientation is now only changed if the gate is actually opened
- You were right, the gate opens itself in the opposite direction the player is facing, but ONLY if the player is facing exactly in this direction. Otherwise the gate is opened to the previous direction. Look at the current code, for more information; should be clear then.

SpaceManiac:
I was trying to change `BlockFenceGate` to use the existing `getOppositeBlockFace` instead of its own `blockFaceFromYaw`, but in doing so noticed this oddity in Bukkit (that was causing this attempt to fail):
https://github.com/GlowstoneMC/Glowkit/blob/master/src/main/java/org/bukkit/material/Gate.java#L29

Any thoughts? Should this be changed Glowkit-side so it isn't so confusing, or left to sit and overridden in BlockFence instead?

Tonodus:
Well, I'd prefer changing the method in Glowkit, so that if we set facingDirection to NORTH it is actually faced to north. The amount of plugins that are broken by this change should be minimal.

socram8888:
@Tonodus: I was also implementing gates when I noticed your PR here. Since this looks prettier than mine, I've only submitted the Glowkit part, which should be useful here: https://github.com/GlowstoneMC/Glowkit/pull/26/files

turt2live:
Bugs from testing this:
- Iron trap doors cannot be placed on the "upper" half of the block. 
  - Don't worry about redstone functionality. But placeholder code for open/close should be there

And one other bug (that is also present in vanilla): 

Opening the fence gate from a perpendicular position then crossing "around" the gate to get to the other side to open on a shallow angle can cause the gate to open the "wrong" way. This is a mathematical error and I don't think it's even worth fixing to be honest. (Graphic included below for a visual representation of what I am talking about)

image

Tonodus:
1a: Added IronTrapDoors
1b: Donno what placeholder code for open/close I should write.. We still can just use the MaterialData Openable to call setOpen()/isOpen().
2: it was harder to implement that bug, than making it "correct". If you do it correct there will be a client side glitch (see comments above). I can change it back to the "right" approach of course, if you prefer that.

turt2live:
As-is, it is the "right" approach as it matches the vanilla server. I just mentioned it so others, if they experience it, don't complain about it. It's a non-issue.

SpaceManiac:
I don't understand what `BlockTrapDoor` is accomplishing here. Why not just have `BlockWoodenTrapDoor` and `BlockIronTrapDoor` _extend_ `BlockTrapDoor` instead of the weirdness you're doing right now?

turt2live:
@Tonodus Your PR is now not mergable due to recent changes. Please update your PR as soon as you can, thanks!

dequis:
Trivial conflict in ItemTable.java, just use both sides.

Merged branch here https://github.com/dequis/Glowstone/compare/pr-200 (which shouldn't even be needed, tbh).

Tonodus:
@SpaceManiac 

> Why not just have BlockWoodenTrapDoor and BlockIronTrapDoor extend BlockTrapDoor

_Short answer:_
Because BlockWoodenTrapDoor extends already BlockOpenable.

_Long answer:_
The current BlockType-system is insufficient. It depends on inheritance instead of composition, what causes the issues as mentioned above: We're running into 'diamond of death's.
Here it's the classes BlockTrapDoor/BlockOpenable, but there are plenty cases where we need composition, as inheritance is insufficient:
F.e. many blocks drop themselves if the block they're attached to is broken. But this is a totally different feature of this block compared to the remain, e.g. block-growth/block-ticking. Both will likely be summarized in superclasses, but the concret class (e.g. BlockSapling) needs both of them.

SpaceManiac:
Merged at long last in 9f19dc9d1ad0b54c589dcb9e6ea09c06a4eae0d0.

PR #195, opened 2014-09-07
Added CONTRIBUTING.md based on original drafts

JeromSar:
This pull request incorporates contributing information based on the [original drafts](https://gist.github.com/SpaceManiac/a5f6d04c4f095c988fcc).

I attempted to carefully describe how a change should proceed. Included is in-depth detail describing what a contributor might expect.

Feedback is greatly appreciated.

PR #201, opened 2014-09-07
Force LF

anna-is-cute:
We need to choose a line-ending format. Personally, and as most people will agree, I believe LF is the best option here. This .gitattributes should convert bad line-endings to LF on push. Along with this PR, I would like to convert all CRLF files to LF, but I haven't included that, as it may break other PRs.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+text eol=lf
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ target
 !.gitignore
 !.gitmodules
 !.mailmap
+!.gitattributes


turt2live:
Please review the CONTRI- oh.

JeromSar:
@turt2live Go go go: https://github.com/SpaceManiac/Glowstone/pull/195

anna-is-cute:
This should enforce #195! :D

ewized:
:+1:

.gitattributes
```diff
@@ -0,0 +1 @@
+text eol=lf
```
turt2live, line 1: This comment is just to annoy @jkcclemens

anna-is-cute, line 1: Thanks.

EntityReborn, line 1: Secondary useless comment of doom.

SpaceManiac:
Merged: 74efe319c929a5bcf7af8392c115e8380be6caac.

Issue #116, opened 2014-06-14
Redstone support


SpaceManiac:
Redstone is not currently supported in any capacity. This should happen at some point.
- [ ] GlowBlock supports redstone methods
- [ ] Redstone wire should transfer current as expected in all cases
- [ ] Blocks which produce redstone current should do so as in Vanilla
- [ ] Blocks which perform actions on redstone input should do so as in Vanilla

iamgreaser:
I'm making a prototype for this, which I hope to transplant into the real thing once it's working. So far there's logic in the prototype but it's broken.

SamNosliw:
Hey, I just wanted to say that I added partial redstone support for 'GlowstonePlusPlus', its not pretty and does need some tidying up, but if you were interested I could transfer it over (hopefully tidier too) to this project. Link: https://github.com/GlowstonePlusPlus/GlowstonePlusPlus/pull/64

turt2live:
@SamNosliw Please make a pull request on our side for review. Thanks!

PR #215, opened 2014-09-08
Redstone framework

iamgreaser:
This is just the start of a redstone framework which actually ended up being somewhat different from the initial prototype.

This is for issue #116 Redstone Support. I have dropped the WIP tag as I am satisfied that I have a suitable working subset of RS functionality (even though the torch timing is 1-tick rather than 2-tick).

Current status:
- **DONE** - RSManager class tied to world to handle the redstone network
- **DONE** - Chunks are dirtied, updated and unloaded by GlowWorld and GlowChunk
- **DONE** - pulse() method hooked up to GlowWorld
- **DONE** - Actually load chunk data into RSManager
- **DONE** - Redstone torch blocks
- **DONE** - Basic redstone logic
  - **DONE** - Redstone torch->solid->torch sequence
  - **DONE** - Redstone wires - diagonals ARE blocked, and these ARE analogue! (NEW: "injection" and charging of solids)
- **FIXED ITSELF?!** - Actually relay redstone updates to client (should be there BUT there's some stupid bug somewhere which is preventing this from actually working)
- **DONE** - Prevent placement of redstone dust ("wires") on a lot of things (mostly RS dust, RS torches)

To be implemented in later pull requests:
- Actually deregister chunks properly
- Faster deregistering of chunks
- Proper network creation which doesn't require tracing wires every pulse (prototype uses HashSets to mark positions of "sources", torches, and charged wires - it's not going to hit _every_ block each frame!)
- Repeaters
- Comparators

### DIFF
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -76,6 +76,11 @@ public final class GlowWorld implements World {
     private final ChunkManager chunks;
 
     /**
+     * The redstone manager.
+     */
+    private final RSManager rsManager;
+
+    /**
      * A lock kept on the spawn chunks.
      */
     private final ChunkManager.ChunkLock spawnChunkLock;
@@ -228,6 +233,7 @@ public final class GlowWorld implements World {
         storageProvider = new AnvilWorldStorageProvider(new File(server.getWorldContainer(), name));
         storageProvider.setWorld(this);
         chunks = new ChunkManager(this, storageProvider.getChunkIoService(), generator);
+        rsManager = new RSManager(this);
         populators = generator.getDefaultPopulators(this);
 
         // set up values from server defaults
@@ -328,6 +334,14 @@ public final class GlowWorld implements World {
     }
 
     /**
+     * Get the world RS manager
+     * @return The RSManager for the world.
+     */
+    public RSManager getRSManager() {
+        return rsManager;
+    }
+
+    /**
      * Get the world's parent server.
      * @return The GlowServer for the world.
      */
@@ -404,6 +418,9 @@ public final class GlowWorld implements World {
                 }
             }
         }
+
+        // tick redstone
+        getRSManager().pulse();
 
         if (--saveTimer <= 0) {
             saveTimer = AUTOSAVE_TIME;
@@ -851,7 +868,9 @@ public final class GlowWorld implements World {
 
     @Override
     public boolean unloadChunk(int x, int z, boolean save, boolean safe) {
-        return !isChunkLoaded(x, z) || getChunkAt(x, z).unload(save, safe);
+        boolean success = (!isChunkLoaded(x, z) || getChunkAt(x, z).unload(save, safe));
+        if(success) { getRSManager().dropRedChunk(x, z); }
+        return success;
     }
 
     @Override
@@ -895,6 +914,8 @@ public final class GlowWorld implements World {
                 result = true;
             }
         }
+
+        getRSManager().dirtyRedChunk(x, z);
 
         return result;
     }
@@ -1335,4 +1356,4 @@ public final class GlowWorld implements World {
         }
         return result;
     }
-}
+} 
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -1358,4 +1358,4 @@ public final class GlowWorld implements World {
         }
         return result;
     }
-} 
+}
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -869,7 +869,9 @@ public final class GlowWorld implements World {
     @Override
     public boolean unloadChunk(int x, int z, boolean save, boolean safe) {
         boolean success = (!isChunkLoaded(x, z) || getChunkAt(x, z).unload(save, safe));
-        if(success) { getRSManager().dropRedChunk(x, z); }
+        if(success) {
+            getRSManager().dropChunk(x, z);
+        }
         return success;
     }
 
@@ -915,7 +917,7 @@ public final class GlowWorld implements World {
             }
         }
 
-        getRSManager().dirtyRedChunk(x, z);
+        getRSManager().dirtyChunk(x, z);
 
         return result;
     }
--- a/src/main/java/net/glowstone/RSManager.java
+++ b/src/main/java/net/glowstone/RSManager.java
@@ -1,0 +1,301 @@
+package net.glowstone;
+
+import net.glowstone.GlowServer;
+import net.glowstone.GlowWorld;
+import java.util.*;
+
+/**
+ * A class to manage redstone logic for a world.
+ * @author Ben Russell
+ */
+public class RSManager
+{
+    /**
+     * A manual enumeration of block types.
+     * TODO: Something better, perhaps?
+     */
+    public final int B_AIR = 0;
+    public final int B_SOLID = 1;
+    public final int B_RWIRE = 2;
+    public final int B_RTORCH = 3;
+    public final int B_RDIODE = 4; // Not supported yet
+
+    /**
+     * An immutable class for storing 3D world positions.
+     * TODO: Consider using Location (assuming it's hashable to the nearest int^3)
+     */
+    public final class RSPos
+    {
+        public final int x, y, z;
+
+        public RSPos(int x, int y, int z)
+        {
+            this.x = x;
+            this.y = y;
+            this.z = z;
+        }
+
+        @Override
+        public int hashCode()
+        {
+            int hash = 1;
+
+            hash = hash * 31 + this.y;
+            hash = hash * 53 + this.x;
+            hash = hash * 19 + this.z;
+
+            return hash;
+        }
+
+        @Override
+        public boolean equals(Object o)
+        {
+            if(o instanceof RSPos)
+            {
+                RSPos op = (RSPos)o;
+
+                return true
+                    && op.x == this.x
+                    && op.y == this.y
+                    && op.z == this.z;
+
+            }
+
+            return false;
+        }
+
+        @Override
+        public String toString()
+        {
+            return String.format("RSPos(%d, %d, %d)", x, y, z);
+        }
+    }
+
+    /**
+     * The world this manager is assigned to.
+     */
+    private GlowWorld world;
+
+    /**
+     * Various redstone-related sets, maps and lists.
+     */ 
+    private HashSet<RSPos> redTorches = new HashSet<RSPos>();
+    private HashSet<RSPos> redSources = new HashSet<RSPos>();
+    private ArrayList<RSPos> redChargedWires = new ArrayList<RSPos>();
+    private HashSet<RSPos> redSinks = new HashSet<RSPos>();
+    private HashMap<RSPos, byte[]> redTypes = new HashMap<RSPos, byte[]>();
+    private HashMap<RSPos, byte[]> redMeta = new HashMap<RSPos, byte[]>();
+    private HashSet<RSPos> chunksDirty = new HashSet<RSPos>();
+    private HashSet<RSPos> chunksDirtyBack = new HashSet<RSPos>();
+
+    /**
+     * Create a new RS manager for a world.
+     * @param world The world this manager pertains to
+     */
+    public RSManager(GlowWorld world)
+    {
+        this.world = world;
+        // TODO: Get a list of chunks and mark the loaded ones as dirty
+    }
+
+    /**
+     * Sets the charge of a given block.
+     * @param x
+     * @param y
+     * @param z
+     * @param charged Whether this block is charged or not
+     */
+    private void setBlockCharge(int x, int y, int z, boolean charged)
+    {
+        // TODO!
+    }
+
+    /**
+     * Drops a redstone chunk out of the network.
+     * @param chunkX
+     * @param chunkZ
+     */
+    public synchronized void dropRedChunk(int chunkX, int chunkZ)
+    {
+        // TODO: Check if we need to prevent double-removals of a chunk
+        RSPos cp = new RSPos(chunkX, 0, chunkZ);
+
+        // Delete redTypes/redMeta
+        redTypes.remove(cp);
+        redMeta .remove(cp);
+
+        //System.out.println(String.format("red chunk !DROP! %d %d", chunkX, chunkZ));
+
+        // Loop
+        for(int y = 0; y < 256; y++)
+        for(int z = 0; z < 16; z++)
+        for(int x = 0; x < 16; x++)
+        {
+            // Get position
+            int rx = x + (chunkX << 4);
+            int ry = y;
+            int rz = z + (chunkZ << 4);
+            RSPos p = new RSPos(rx, ry, rz);
+
+            // Remove this from all appropriate sets
+            redTorches.remove(p);
+            redSources.remove(p);
+            redSinks.remove(p);
+        }
+    }
+
+    /**
+     * Marks a chunk as dirty so the redstone manager can update it.
+     * @param chunkX
+     * @param chunkZ
+     */
+    public synchronized void dirtyRedChunk(int chunkX, int chunkZ)
+    {
+        chunksDirty.add(new RSPos(chunkX, 0, chunkZ));
+    }
+
+    /**
+     * Updates the redstone state for a chunk.
+     * @param chunkX
+     * @param chunkZ
+     */
+    private synchronized void updateRedChunk(int chunkX, int chunkZ)
+    {
+        //System.out.println(String.format("red chunk update %d %d", chunkX, chunkZ));
+        RSPos cp = new RSPos(chunkX, 0, chunkZ);
+
+        // Fetch IDs and metadata
+        byte[] blockIDs  = (redTypes.containsKey(cp) ? redTypes.get(cp) : new byte[16*16*256]);
+        byte[] blockMeta = (redMeta .containsKey(cp) ? redMeta .get(cp) : new byte[16*16*256]);
+
+        // Loop
+        for(int pos = 0, y = 0; y < 256; y++)
+        for(int z = 0; z < 16; z++)
+        for(int x = 0; x < 16; x++, pos++)
+        {
+            // Get type from chunk (TODO!)
+            int type = B_AIR;
+            int meta = 0x00;
+            boolean charged = false;
+
+            // Write type
+            blockIDs[pos]  = (byte)type;
+            blockMeta[pos] = (byte)meta;
+
+            // Get position
+            int rx = x + (chunkX << 4);
+            int ry = y;
+            int rz = z + (chunkZ << 4);
+            RSPos p = new RSPos(rx, ry, rz);
+
+            // Remove this from all appropriate sets
+            redTorches.remove(p);
+            redSources.remove(p);
+            redSinks.remove(p);
+
+            switch(type)
+            {
+                case B_AIR:
+                    // Air is never a source
+                    break;
+
+                case B_RWIRE:
+                    // Wire isn't actually used at this step
+                    break;
+
+                case B_RTORCH:
+                    // Add this to the torches AND charges AND sources
+                    p = new RSPos(rx, ry, rz);
+                    redTorches.add(p);
+                    if(!charged) redSources.add(p);
+                    break;
+
+                case B_SOLID:
+                    // Add this to the charges AND sources
+                    p = new RSPos(rx, ry, rz);
+                    if(charged) redSources.add(p);
+                    break;
+
+            }
+        }
+
+        // Update redTypes and redMeta
+        redTypes.put(cp, blockIDs );
+        redMeta .put(cp, blockMeta);
+    }
+
+    /**
+     * Determine if a given chunk is loaded or not.
+     */
+    private boolean isChunkLoaded(int x, int z)
+    {
+        return world.isChunkLoaded(x, z);
+    }
+
+    /**
+     * Update all the redstone logic in the manager's assigned world.
+     */
+    public synchronized void pulse()
+    {
+        // Swap chunksDirty buffers
+        HashSet<RSPos> cdtemp = chunksDirty;
+        chunksDirty = chunksDirtyBack;
+        chunksDirtyBack = cdtemp;
+
+        // Clear dirty chunks
+        chunksDirty.clear();
+
+        // Update all dirty chunks
+        for(RSPos p : chunksDirtyBack)
+        {
+            // Skip unloaded chunks
+            if(!isChunkLoaded(p.x, p.z)) continue;
+
+            // Update RS chunk
+            updateRedChunk(p.x, p.z);
+        }
+
+        // Clear sinks
+        redSinks.clear();
+
+        // Discharge wires
+        for(RSPos p : redChargedWires)
+            setBlockCharge(p.x, p.y, p.z, false);
+
+        redChargedWires.clear();
+
+        // Lead sources to sinks
+        // TODO: Integrate properly
+        /*
+        for(RSPos p : redSources)
+            traceRedStart(p.x, p.y, p.z);
+        */
+
+        // Discharge sources
+        for(RSPos p : redSources)
+            setBlockCharge(p.x, p.y, p.z, false);
+
+        // Charge sinks
+        for(RSPos p : redSinks)
+            setBlockCharge(p.x, p.y, p.z, true);
+
+        // Charge wires
+        for(RSPos p : redChargedWires)
+            setBlockCharge(p.x, p.y, p.z, true);
+
+        // Flip sinks for torches
+        for(RSPos p : redTorches)
+        {
+            if(redSinks.contains(p))
+                redSinks.remove(p);
+            else
+                redSinks.add(p);
+        }
+
+        // Swap RS sources and sinks
+        HashSet<RSPos> rstemp = redSources;
+        redSources = redSinks;
+        redSinks = rstemp;
+    }
+}
+
--- a/src/main/java/net/glowstone/RSManager.java
+++ b/src/main/java/net/glowstone/RSManager.java
@@ -77,8 +77,10 @@ public class RSManager {
     /**
      * Various redstone-related sets, maps and lists.
      */ 
-    private Map<RSPos,Integer> redChargeNew = new HashMap<RSPos,Integer>();
-    private Map<RSPos,Integer> redChargeOld = new HashMap<RSPos,Integer>();
+    private Map<RSPos,Integer> redPowerNew = new HashMap<RSPos,Integer>();
+    private Map<RSPos,Integer> redPowerOld = new HashMap<RSPos,Integer>();
+    private Map<RSPos,Integer> redPowerRead = redPowerNew;
+    private Set<RSPos> redPowerFlush = new HashSet<RSPos>();
     private Set<RSPos> redSourceNew = new HashSet<RSPos>();
     private Set<RSPos> redSourceOld = new HashSet<RSPos>();
     private Set<RSPos> chunksDirtyNew = new HashSet<RSPos>();
@@ -103,10 +105,10 @@ public class RSManager {
      * @param charge The charge level to set this block to.
      * @param oneTick Flag to indicate if we must discharge this block in the next tick.
      */
-    public void setBlockCharge(int x, int y, int z, int charge, boolean oneTick) {
+    public synchronized void setBlockPower(int x, int y, int z, int charge, boolean oneTick) {
         assert(charge >= 0 && charge <= 15);
         RSPos p = new RSPos(x, y, z);
-        redChargeNew.put(p, (Integer)((charge&15) | (oneTick?0x20:0x00)));
+        redPowerNew.put(p, (Integer)((charge&15) | (oneTick?0x20:0x00)));
     }
 
     /**
@@ -115,8 +117,8 @@ public class RSManager {
      * @param charge The charge level to set this block to.
      * @param oneTick Flag to indicate if we must discharge this block in the next tick.
      */
-    public void setBlockCharge(Block block, int charge, boolean oneTick) {
-        setBlockCharge(block.getX(), block.getY(), block.getZ(), charge, oneTick);
+    public void setBlockPower(Block block, int charge, boolean oneTick) {
+        setBlockPower(block.getX(), block.getY(), block.getZ(), charge, oneTick);
     }
 
     /**
@@ -125,9 +127,54 @@ public class RSManager {
      * @param charge The charge level to set this block to.
      * @param oneTick Flag to indicate if we must discharge this block in the next tick.
      */
-    public void setBlockCharge(RSPos p, int charge, boolean oneTick) {
-        setBlockCharge(p.x, p.y, p.z, charge, oneTick);
+    public void setBlockPower(RSPos p, int charge, boolean oneTick) {
+        setBlockPower(p.x, p.y, p.z, charge, oneTick);
     }
+
+    /**
+     * Gets the charge of a given block.
+     * @param x
+     * @param y
+     * @param z
+     * @return Power level in the range [0, 15].
+     */
+    public synchronized int getBlockPower(int x, int y, int z) {
+        RSPos p = new RSPos(x, y, z);
+        Integer charge = redPowerRead.get(p);
+        return (charge == null ? 0 : ((int)charge) & 15);
+    }
+
+    /**
+     * Gets the charge of a given block.
+     * @param block The block to get the charge of.
+     * @return Power level in the range [0, 15].
+     */
+    public int getBlockPower(Block block) {
+        return getBlockPower(block.getX(), block.getY(), block.getZ());
+    }
+
+    /**
+     * Gets the to-be-applied charge of a given block.
+     * @param x
+     * @param y
+     * @param z
+     * @return Power level in the range [0, 15].
+     */
+    public synchronized int getNewBlockPower(int x, int y, int z) {
+        RSPos p = new RSPos(x, y, z);
+        Integer charge = redPowerNew.get(p);
+        return (charge == null ? 0 : ((int)charge) & 15);
+    }
+
+    /**
+     * Gets the to-be-applied charge of a given block.
+     * @param block The block to get the charge of.
+     * @return Power level in the range [0, 15].
+     */
+    public int getNewBlockPower(Block block) {
+        return getNewBlockPower(block.getX(), block.getY(), block.getZ());
+    }
+
 
     /**
      * Trace from a block to another block.
@@ -175,7 +222,7 @@ public class RSManager {
      * Adds a source for tracing updates.
      * @param block The block we use as a source.
      */
-    public void addSource(Block block) {
+    public synchronized void addSource(Block block) {
         RSPos p = new RSPos(block);
         redSourceNew.add(p);
         if(DEBUG_REDSTONE) {
@@ -207,7 +254,7 @@ public class RSManager {
             RSPos p = new RSPos(rx, ry, rz);
 
             // Remove this from all appropriate sets
-            redChargeNew.remove(p);
+            redPowerNew.remove(p);
             redSourceNew.remove(p);
         }
     }
@@ -346,10 +393,10 @@ public class RSManager {
         redSourceNew.clear();
 
         // Swap charges and clear new
-        Map<RSPos,Integer> redChargeTemp = redChargeNew;
-        redChargeNew = redChargeOld;
-        redChargeOld = redChargeTemp;
-        redChargeNew.clear();
+        Map<RSPos,Integer> redPowerTemp = redPowerNew;
+        redPowerNew = redPowerOld;
+        redPowerOld = redPowerTemp;
+        redPowerNew.clear();
 
         // Initialise sources
         for(RSPos p : redSourceOld) {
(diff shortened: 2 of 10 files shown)


src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -326,6 +332,14 @@ public ChunkManager getChunkManager() {
     }
 
     /**
+     * Get the world RS manager
```
turt2live: Missing period.

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -782,7 +799,9 @@ public boolean unloadChunk(int x, int z, boolean save) {
     }
 
     public boolean unloadChunk(int x, int z, boolean save, boolean safe) {
-        return !isChunkLoaded(x, z) || getChunkAt(x, z).unload(save, safe);
+        boolean success = !isChunkLoaded(x, z) || getChunkAt(x, z).unload(save, safe);
+        if(success) rsManager.dropRedChunk(x, z);
```
turt2live: Conventions say to use `{` and `}`

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,301 @@
...
+import java.util.*;
+
+/**
+ * A class to manage redstone logic for a world.
+ * @author Ben Russell
+ */
+public class RSManager
+{
+    /**
+     * A manual enumeration of block types.
```
turt2live: A proper enumeration may be better.

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,301 @@
...
+     */
+    public final int B_AIR = 0;
+    public final int B_SOLID = 1;
+    public final int B_RWIRE = 2;
+    public final int B_RTORCH = 3;
+    public final int B_RDIODE = 4; // Not supported yet
+
+    /**
+     * An immutable class for storing 3D world positions.
+     * TODO: Consider using Location (assuming it's hashable to the nearest int^3)
```
turt2live: It should be. I don't see a problem with it.

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,301 @@
...
+            redTorches.remove(p);
+            redSources.remove(p);
+            redSinks.remove(p);
+
+            switch(type)
+            {
+                case B_AIR:
+                    // Air is never a source
+                    break;
+
```
turt2live: Whitespace

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,301 @@
...
+                case B_RWIRE:
+                    // Wire isn't actually used at this step
+                    break;
+
```
turt2live: Whitespace

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,301 @@
...
+                case B_RTORCH:
+                    // Add this to the torches AND charges AND sources
+                    p = new RSPos(rx, ry, rz);
+                    redTorches.add(p);
+                    if(!charged) redSources.add(p);
+                    break;
+
```
turt2live: Whitespace

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,301 @@
...
+                case B_SOLID:
+                    // Add this to the charges AND sources
+                    p = new RSPos(rx, ry, rz);
+                    if(charged) redSources.add(p);
+                    break;
+
```
turt2live: Whitespace

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,301 @@
...
+                redSinks.add(p);
+        }
+
+        // Swap RS sources and sinks
+        HashSet<RSPos> rstemp = redSources;
+        redSources = redSinks;
+        redSinks = rstemp;
+    }
+}
+
```
turt2live: Whitespace

src/main/java/net/glowstone/block/blocktype/BlockTorch.java
```diff
@@ -0,0 +1,46 @@
...
+import net.glowstone.block.GlowBlockState;
+import net.glowstone.entity.GlowPlayer;
+import org.bukkit.Material;
+import org.bukkit.block.BlockFace;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.util.Vector;
+
+import java.util.*;
+
+public class BlockTorch extends BlockType {
```
turt2live: Blank line afterwards

src/main/java/net/glowstone/block/blocktype/BlockTorch.java
```diff
@@ -0,0 +1,46 @@
...
+        state.setType(matType);
+        state.setRawData((byte)getFacing(face));
+    }
+
+    @Override
+    public Collection<ItemStack> getDrops(GlowBlock block) {
+        return Collections.unmodifiableList(Arrays.asList(new ItemStack(matType, 1, (byte)0)));
+    }
+}
+
```
turt2live: Whitespace

turt2live:
@iamgreaser In addition to the comments made, there are also hanging braces.

The style should be
`if (condition) {`
`public void method(parameters) {`

but instead there are many places where a hanging brace occurs, like below.

```
if (condition) 
{
```

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,301 @@
...
+     * Various redstone-related sets, maps and lists.
+     */ 
+    private HashSet<RSPos> redTorches = new HashSet<RSPos>();
+    private HashSet<RSPos> redSources = new HashSet<RSPos>();
+    private ArrayList<RSPos> redChargedWires = new ArrayList<RSPos>();
+    private HashSet<RSPos> redSinks = new HashSet<RSPos>();
+    private HashMap<RSPos, byte[]> redTypes = new HashMap<RSPos, byte[]>();
+    private HashMap<RSPos, byte[]> redMeta = new HashMap<RSPos, byte[]>();
+    private HashSet<RSPos> chunksDirty = new HashSet<RSPos>();
+    private HashSet<RSPos> chunksDirtyBack = new HashSet<RSPos>();
```
Ribesg: Should use `Set` as variable type, not `HashSet`. Same for everything above this line

turt2live: Also, if possible, find a way to not store every redstone device known to man in the class.

src/main/java/net/glowstone/GlowWorld.java
iamgreaser: A few lines below I see:

```
    public boolean unloadChunkRequest(final int x, final int z, final boolean safe) {
        if (safe && isChunkInUse(x, z)) return false;
```

Guess I should also update that?

src/main/java/net/glowstone/RSManager.java
iamgreaser:
```
:%g/^\s*{\s*$/-1join
```

Next.

src/main/java/net/glowstone/GlowWorld.java
gdude2002: Plis.

Seriously though, always best to use `{ }` for conditionals. Very easy to introduce nasty unintentional bugs.

src/main/java/net/glowstone/RSManager.java
iamgreaser: Which lines are we talking about?

When I'm editing I tend to have whitespace between cases because it's much more convenient when I'm in VIM - it means I can use the { and } keys to move between cases.

I've packed the cases together now. That could be something worth looking at, though.

src/main/java/net/glowstone/block/blocktype/BlockTorch.java
iamgreaser: I really don't know what you're getting at, unless you're talking about the empty line at the end.

src/main/java/net/glowstone/RSManager.java
iamgreaser: I'll sort out that "store every redstone device" thing in time.

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,485 @@
...
+        @Override
+        public String toString() {
+            return String.format("RSPos(%d, %d, %d)", x, y, z);
+        }
+    }
+
+    /**
+     * The world this manager is assigned to.
+     */
+    private GlowWorld world;
```
EntityReborn, line 75: Stuff like this should go at the top of your class.

iamgreaser, line 75: The DEBUG_REDSTONE flag was placed there deliberately to make it easy to access just in case I or someone needs to enable it.

Or are you saying that I should move the other variables up above the inner class?

EntityReborn, line 75: Oh, pfft. Didn't even notice it was an inner-class. Nevermind.

src/main/java/net/glowstone/block/blocktype/BlockTorch.java
```diff
@@ -0,0 +1,163 @@
...
+        Material thisMat = getMaterial();
+        return thisMat == Material.REDSTONE_TORCH_ON || thisMat == Material.REDSTONE_TORCH_OFF;
+    }
+
+    @Override
+    public void traceBlockPowerEnd(GlowBlock block, RSManager rsManager, int power) {
+        // Handle power change.
+        // TODO: make this actually work
+        Material mat = block.getType();
+        if(mat == Material.REDSTONE_TORCH_ON && power == 0) {
```
EntityReborn: You'll want to make a call to GlowServer.getScheduler().scheduleDyncDelatedTask() to do this code here, to implement the 2 tick behavior. That is, if I understand your code correctly.

src/main/java/net/glowstone/block/blocktype/BlockTorch.java
```diff
@@ -0,0 +1,163 @@
...
+
+        // Trace outwards.
+        rsManager.traceFromBlock(srcBlock, toDir, 15, isDirect);
+    }
+
+    @Override
+    public void traceBlockPowerInit(GlowBlock block, RSManager rsManager) {
+        Material thisMat = getMaterial();
+        if(thisMat == Material.REDSTONE_TORCH_ON) {
+            rsManager.setBlockPower(block, 15, false);
```
EntityReborn: This code is exactly the same as line 114, oops?

src/main/java/net/glowstone/block/blocktype/BlockRedstoneDust.java
```diff
@@ -0,0 +1,190 @@
...
+        if(nn) { bsum++; }
+        if(ns) { bsum++; }
+        if(ne) { bsum++; }
+        if(nw) { bsum++; }
+
+        // Flow behaves differently from top and bottom
+        if(flowDir == BlockFace.UP || flowDir == BlockFace.DOWN) {
+            if(bsum == 0) {
+                // If there are no wires, "inject" in a spread
+                traceBlockPowerInject(block, rsManager, BlockFace.NORTH, outPower);
```
EntityReborn, line 145: These code blocks smell like they could be refactored. Possibly send in an array of BlockFaces?

src/main/java/net/glowstone/block/blocktype/BlockTorch.java
```diff
@@ -0,0 +1,163 @@
...
+        // Abandon ship if this is not a charged RS torch.
+        Material thisMat = getMaterial();
+        if(thisMat != Material.REDSTONE_TORCH_ON) {
+            return;
+        }
+
+        // Trace in all directions except self.
+        BlockFace selfDir = getOwnFacing(block).getOppositeFace();
+        traceBlockPowerFromRSTorch(block, rsManager, selfDir, BlockFace.UP, true);
+        traceBlockPowerFromRSTorch(block, rsManager, selfDir, BlockFace.DOWN, false);
```
EntityReborn: These code blocks smell like they could be refactored. Possibly send in an array of BlockFaces?

src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -192,6 +194,155 @@ public final void rightClickBlock(GlowPlayer player, GlowBlock against, BlockFac
...
+    public void traceBlockPowerStart(GlowBlock block, RSManager rsManager) {
+        // TODO: Make note of this block's "charge" level
+        // (will be necessary once we get onto implementing repeaters)
+
+        // Check if solid
+        if(block.getType() == null || !block.getType().isOccluding()) {
+            return;
+        }
+
+        traceBlockPowerStartSolid(block, rsManager, BlockFace.UP, false);
```
EntityReborn, line 301: These code blocks smell like they could be refactored. Possibly send in an array of BlockFaces?

src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -192,6 +194,155 @@ public final void rightClickBlock(GlowPlayer player, GlowBlock against, BlockFac
...
+            rsManager.addSource(block);
+        } else {
+            if(srcMat != Material.REDSTONE_WIRE) {
+                return;
+            }
+        }
+
+        // Spread to neighbours
+        BlockFace oppDir = flowDir.getOppositeFace();
+        traceBlockPowerSolidToBlock(block, rsManager, oppDir, BlockFace.UP, isDirect);
```
EntityReborn, line 335: These code blocks smell like they could be refactored. Possibly send in an array of BlockFaces?

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,514 @@
...
+/**
+ * A class to manage redstone logic for a world.
+ * @author Ben Russell
+ */
+public class RSManager {
+    /**
+     * A flag used to enable testing redstone by spamming the console.
+     * Best to only enable this if you are a developer.
+     */
+    public static final boolean DEBUG_REDSTONE = false;
```
turt2live, line 22: Not sure if this is required. IDEs have breakpoints.

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,514 @@
...
+
+    /**
+     * An immutable class for storing 3D world positions.
+     */
+    public final class RSPos {
```
turt2live, line 27: This really shouldn't be here. Please find an alternative or move it to a more appropriate package.

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,514 @@
...
+
+    /**
+     * The world this manager is assigned to.
+     */
+    private GlowWorld world;
+
+    /**
+     * Various redstone-related sets, maps and lists.
+     */ 
+    private Map<RSPos,Integer> redPowerNew = new HashMap<RSPos,Integer>();
```
turt2live, line 80: I do honestly challenge the idea of a set/map for every redstone device in the world.

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,514 @@
...
+    private Set<RSPos> blocksDirtyNew = new HashSet<RSPos>();
+    private Set<RSPos> blocksDirtyOld = new HashSet<RSPos>();
+
+    /**
+     * Create a new RS manager for a world.
+     * @param world The world this manager pertains to
+     */
+    public RSManager(GlowWorld world) {
+        this.world = world;
+        // TODO: Get a list of chunks and mark the loaded ones as dirty
```
turt2live: Should be done/removed before the PR is pulled.

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,514 @@
...
+    /**
+     * Gets the to-be-applied charge of a given block.
+     * @param block The block to get the charge of.
+     * @return Power level in the range [0, 15].
+     */
+    public int getNewBlockPower(Block block) {
+        return getNewBlockPower(block.getX(), block.getY(), block.getZ());
+    }
+
+
```
turt2live: Whitespace

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,514 @@
...
+    /**
+     * Trace to a block from another block.
+     * @param srcBlock The block we are sourcing from.
+     * @param destBlock The block we are targeting.
+     * @param flowDir The direction of redstone flow from the source block.
+     * @param inPower The input power level.
+     * @param isDirect Whether we are applying direct or indirect power.
+     */
+    public void traceFromBlockToBlock(GlowBlock srcBlock, GlowBlock destBlock, BlockFace flowDir, int inPower, boolean isDirect) {
```
turt2live, line 185: These trace methods aren't very clear. They don't return anything, but seem to do something with respect to tracing. The documentation isn't clear either.

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,514 @@
...
+    public synchronized void dirtyRedBlock(int x, int y, int z) {
+        blocksDirtyNew.add(new RSPos(x, y, z));
+    }
+
+    /**
+     * Marks a chunk as dirty so the redstone manager can update it.
+     * @param chunkX
+     * @param chunkZ
+     */
+    public synchronized void dirtyRedChunk(int chunkX, int chunkZ) {
```
turt2live: Considering the class name, I don't think `Red` is required in this (or similar) method names.

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,514 @@
...
+    private synchronized void updateRedChunk(int chunkX, int chunkZ) {
+        if(DEBUG_REDSTONE) {
+            //System.out.println(String.format("red chunk update %d %d", chunkX, chunkZ));
+        }
+
+        // Fetch chunk
+        GlowChunk chunk = world.getChunkAt(chunkX, chunkZ);
+
+        // Loop
+        for(int pos = 0, y = 0; y < 256; y++)
```
turt2live: You are missing braces on these 2 loops.

src/main/java/net/glowstone/block/GlowBlock.java
```diff
@@ -149,7 +150,8 @@ public GlowBlock getRelative(BlockFace face, int distance) {
     // Type and typeid getters/setters
 
     public Material getType() {
-        return Material.getMaterial(getTypeId());
+        Material mat = Material.getMaterial(getTypeId());
+        return (mat == null ? Material.AIR : mat);
```
turt2live: Something that is an unknown type is not air.

src/main/java/net/glowstone/block/blocktype/BlockRedstoneDust.java
```diff
@@ -0,0 +1,190 @@
...
+import org.bukkit.block.BlockFace;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.ItemStack;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+
+public class BlockRedstoneDust extends BlockType {
+    // TODO: handle properly
```
turt2live: This needs to be fixed before the PR can be pulled.

src/main/java/net/glowstone/block/blocktype/BlockRedstoneDust.java
```diff
@@ -0,0 +1,190 @@
...
+        if(isDirect) {
+            return false;
+        }
+
+        return true;
+    }
+
+    @Override
+    public void traceBlockPowerInit(GlowBlock block, RSManager rsManager) {
+        // XXX: Do we discharge here, or rely on rsManager?
```
turt2live: This should be fixed before this PR can be pulled.

src/main/java/net/glowstone/block/blocktype/BlockRedstoneDust.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    @Override
+    public void traceBlockPowerStart(GlowBlock block, RSManager rsManager) {
+        // This does not trace from the "start".
```
turt2live: This doesn't do anything, actually.

src/main/java/net/glowstone/block/blocktype/BlockTorch.java
```diff
@@ -0,0 +1,163 @@
...
+    @Override
+    public boolean isRedSource(GlowBlock block) {
+        Material thisMat = getMaterial();
+        return thisMat == Material.REDSTONE_TORCH_ON || thisMat == Material.REDSTONE_TORCH_OFF;
+    }
+
+    @Override
+    public void traceBlockPowerEnd(GlowBlock block, RSManager rsManager, int power) {
+        // Handle power change.
+        // TODO: make this actually work
```
turt2live: This needs to be fixed before the PR can be pulled.

src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -192,6 +194,155 @@ public final void rightClickBlock(GlowPlayer player, GlowBlock against, BlockFac
...
+    // Redstone
+
+    /**
+     * Determines whether redstone is emitted from this face or not.
+     * @param block The block we are operating on.
+     * @param face The face of the block that is being touched, or SELF for internal charge.
+     * @param isDirect Whether we are looking for direct or indirect power.
+     */
+    public boolean isBlockEmittingPower(GlowBlock block, BlockFace face, boolean isDirect) {
+        return true;
```
turt2live: All blocks emit power?

src/main/java/net/glowstone/block/blocktype/BlockRedstoneDust.java
```diff
@@ -0,0 +1,190 @@
...
+        // Bail out early
+        if(blockDir == outDir) {
+            return false;
+        }
+        if(outPower <= 0) {
+            return false;
+        }
+
+        // Get the relevant blocks + materials
+        GlowBlock blockOn   = block.getRelative(BlockFace.UP);
```
EntityReborn: The variable names you use here are really confusing.

src/main/java/net/glowstone/block/GlowBlock.java
iamgreaser: Then it should return a dummy ("UNKNOWN", perhaps?) - otherwise you end up having to compensate for NullPointerExceptions all over the code.

turt2live: If there's a NPE, then someone is using an unsupported build or unsupported world.

src/main/java/net/glowstone/block/blocktype/BlockType.java
iamgreaser, line 301: I find it's a lot cleaner to produce the cases like that than having to clutter the code with an array and a for() loop.

turt2live, line 301: This is clutter as-is

EntityReborn, line 301: By passing in a list of some sort, in a lot of cases of repeated code, the end result can me more optimized when dealing with checking and activating functionality.

iamgreaser, line 301: I'm not aiming for optimised at this stage. I'm aiming for ease of modification.

EntityReborn, line 301: Understandable, but be aware, "I'll fix this later" more than not turns into forgetting and never fixing. Do something correctly now, before you regret it (or forget it) later.

src/main/java/net/glowstone/block/GlowBlock.java
iamgreaser: At this stage it's possible to turn a supported world into an unsupported world pretty easily and break the damn server. Until everything has a proper implementation, something like this will have to do.

So, what block should it be?

turt2live: Honestly, that change needs to be made in it's own PR to not break the server. This change is not suitable nor is it suitable in this PR.

src/main/java/net/glowstone/block/blocktype/BlockRedstoneDust.java
iamgreaser: Exactly, that's why it says "this does not trace from the start".

src/main/java/net/glowstone/block/blocktype/BlockType.java
iamgreaser: Should be "canBlockEmitPower". Also should return getMaterial().isOccluded(). I'll sort that out.

src/main/java/net/glowstone/block/blocktype/BlockRedstoneDust.java
iamgreaser: How so? Please give examples.

EntityReborn: For example, you have a blockDown variable, yet blockSeat gets the block at BlockFace.DOWN. Ditto with blockUp and blockOn.

iamgreaser: Propose a better set of names and I'll chuck them in.

src/main/java/net/glowstone/block/blocktype/BlockType.java
turt2live, line 301: I would suggest fixing this as soon as possible

src/main/java/net/glowstone/block/blocktype/BlockRedstoneDust.java
turt2live: That doesn't make any sense when reading the code. It's just a comment that means nothing.

src/main/java/net/glowstone/block/blocktype/BlockRedstoneDust.java
turt2live: GlowBlock[] relatives = new GlowBlock[5]; // Should be whatever value is required. 
// Insert `for` loop for directions.

src/main/java/net/glowstone/block/blocktype/BlockType.java
iamgreaser, line 301: I'd have to actually profile the code to see if that would improve anything. Personally I have no issues with the way it's done right now.

src/main/java/net/glowstone/block/blocktype/BlockRedstoneDust.java
iamgreaser: That makes it more confusing.

src/main/java/net/glowstone/RSManager.java
iamgreaser, line 80: Ultimately there _will_ be a few Sets _somewhere_ so we know what blocks to update. The old and new charges _must_ be tracked somehow, too.

What would be a better way to arrange the data?

src/main/java/net/glowstone/RSManager.java
iamgreaser, line 27: c.f. GlowChunk.Key. This Key class only handles two dimensions, though, so we can't use it here.

org.bukkit.Location is a bit inappropriate because it uses floating point numbers, not integers.

iamgreaser:
And if you ever think about putting a CRLF file in the repo again, I will kill you to death.

src/main/java/net/glowstone/RSManager.java
turt2live, line 27: So move this out of here and to somewhere more repurposable.

src/main/java/net/glowstone/RSManager.java
turt2live, line 80: My concern is overtime Mojang adding more redstone devices. Having a class with a page of field declarations is not at all ideal.

src/main/java/net/glowstone/block/blocktype/BlockType.java
turt2live, line 301: It's a bad coding practice. You can shorten this code to a 7 line reusable method for others to use.

src/main/java/net/glowstone/block/blocktype/BlockRedstoneDust.java
turt2live: It is less confusing if you get a proper utility method setup. See previous comments.

src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,529 @@
...
+    }
+
+    /**
+     * Drops a redstone chunk out of the network.
+     * @param chunkX
+     * @param chunkZ
+     */
+    public synchronized void dropChunk(int chunkX, int chunkZ) {
+        // We could prevent double-removals of a chunk,
+        //
```
turt2live: Blank comment

src/main/java/net/glowstone/block/GlowBlock.java
```diff
@@ -261,32 +262,48 @@ public byte getLightFromBlocks() {
...
     }
 
     @Override
     public boolean isBlockFacePowered(BlockFace face) {
-        throw new UnsupportedOperationException("Not supported yet.");
+        BlockType type = ItemTable.instance().getBlock(getType());
+        if(type == null) {
+            return false;
+        }
+        return type.canBlockEmitPower(this, face, true) &&
```
turt2live: I thought we decided on no "weird" 80-column limit?

src/main/java/net/glowstone/block/GlowBlock.java
```diff
@@ -261,32 +262,48 @@ public byte getLightFromBlocks() {
...
     }
 
     @Override
     public boolean isBlockFaceIndirectlyPowered(BlockFace face) {
-        throw new UnsupportedOperationException("Not supported yet.");
+        BlockType type = ItemTable.instance().getBlock(getType());
+        if(type == null) {
+            return false;
+        }
+        return type.canBlockEmitPower(this, face, false) &&
```
turt2live: https://github.com/GlowstoneMC/Glowstone/pull/215/files#r17402180

src/main/java/net/glowstone/block/GlowBlock.java
```diff
@@ -261,32 +262,48 @@ public byte getLightFromBlocks() {
...
     }
 
     @Override
     public int getBlockPower(BlockFace face) {
-        throw new UnsupportedOperationException("Not supported yet.");
+        BlockType type = ItemTable.instance().getBlock(getType());
+        if(type == null) {
+            return 0;
+        }
+        return (type.canBlockEmitPower(this, face, false)
```
turt2live: https://github.com/GlowstoneMC/Glowstone/pull/215/files#r17402180

src/main/java/net/glowstone/RSManager.java
iamgreaser: Ah whoops, I'll sort that one out.

src/main/java/net/glowstone/block/GlowBlock.java
iamgreaser: I'll leave this one as-is as it makes the ?: a bit clearer.

turt2live: It doesn't fit the style guidelines. Also the parenthesis need to be removed.

ZachBora:
It seems the class BlockTorch contains too many stuff related to Redstone torches. You should make a BlockRedstoneTorch that extends BlockTorch instead.

turt2live:
Sorry - bot fail :(

turt2live:
@iamgreaser There are many cases not currently handled with this PR, such as some redstone clocks that should 'fizzle out', don't. Additionally, I would personally like to see this PR include more than the torches and wire. All redstone devices would be neat to have in one PR.

One example is the following (follow the steps):
1. Create the following: http://i.turt2live.com/9b8G.jpg
2. Add a loop: http://i.turt2live.com/iEcw.jpg
3. Cut off the redstone source: http://i.turt2live.com/Vz09.jpg
4. Watch the clock freak out and send ~100000x more packets than Vanilla and never fizzle out.

src/main/java/net/glowstone/block/blocktype/BlockTorch.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public void placeBlock(GlowPlayer player, GlowBlockState state, BlockFace face, ItemStack holding, Vector clickedLoc) {
+        state.setType(matType);
+        state.setRawData((byte)getFacing(face));
+    }
+
+    @Override
+    public Collection<ItemStack> getDrops(GlowBlock block) {
+        return Collections.unmodifiableList(Arrays.asList(new ItemStack(matType, 1, (byte)0)));
```
ZachBora: I don't see the point in overriding the base class on this one. By default the drops should be the torch that was created (normal or redstone).

anna-is-cute: As was pointed out to me, better to use setDrops in the constructor than override getDrops.

ZachBora: This getdrops will return the same value as the base class, no usage for it here. Setdrop isn't useful either for this particular item. It might be useful in the Redstone torch class so you don't accidently give the user an unlit redstone torch. But this is the torch class.

scooterboo:
@turt2live , @iamgreaser  was having problems rebasing the code because of the sheer size of the PR. I believe repeaters, comparators , pistons, levers, buttons, pressure plates, droppers, hoppers, dispensers,daylight sensors, redstone blocks, and all the little bugs can be fixed and added after this is comitted.

turt2live:
Well, at the very least the bugs with this PR need to be resolved. I suppose the sheer size would be daunting for all devices.

iamgreaser:
@turt2live The fizzling out will not be tackled until another PR, for two reasons:
1. This PR is already very large and I have to limit the scope.
2. I've seen a few of those fast pulsars on my server and it's not a major bandwidth issue.

I'm sorting out the merge now.

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1325,4 +1348,4 @@ public void sendPluginMessage(Plugin source, String channel, byte[] message) {
         }
         return result;
     }
-}
+}
```
daboross: Whitespace?

iamgreaser:
If you aren't happy with the fact that torches do not fizzle out, at this stage I'll have to make redstone support a config option.

Wild1145:
Just so you know, this PR cant be merged at the current time.

iamgreaser:
Closing as this will never get pulled. If someone wants to prove me wrong, please fork it and open a new PR - I encourage you, because if I am wrong, and an RS framework _does_ get pulled, then we all benefit.

While we're at it, here's a probably incomplete list of "bugs" in the code:
- Torches are 1-tick, not 2-tick
- Torches turn on/off at the end of a tick, rather than the start of the next applicable tick
- Torches are placed in the off state to prevent a particular flicker situation
- Torches can be placed on the side of non-occluded blocks
- Wire cannot be placed on the top of an upside-down slab, nor does it behave like the upside-down slabs are glowstone
- Only wire can "stray" a wire away from "injecting" into a block (an RS torch also has the ability)
- getBlockPower() almost definitely returns the wrong values for RS torches at the very least

Oh, and
- Torches don't burn out, raising the bandwidth usage to a whopping... not that much, really.

This is of course ignoring the fact that there aren't any physics to create a drop when you remove the block that a thing is on, but I believe you're well aware that it's a matter reserved for a different pull request.

The main reason I wanted this pushed at its incomplete state was so that more people could work on bringing it to "completion" (ha, like that ever happens when you clone something) faster. Having an incomplete thing is better than having nothing at all, especially when you consider that at this stage Glowstone is "not production-ready" (as much as I hate to say it - I've had that used against my code by users).

**Insisting that this passes _all_ RS tests before it can be pulled is completely unrealistic and completely undermines the point of using version control.**

While I accept that people use version control differently, there's no point in having exactly one branch that's kept as stable as possible when Git allows you to use several branches. Seriously, go add a "stable" branch and use "master" for the rough stuff. Or maybe add a "testing" and/or "unstable" branch. Different people have different requirements, and I'd be more than happy to use a bleeding-edge version even if it's full of bugs - aren't you glad PRs are a thing?

Sure, you may want to play the "experience" card. But having experience in one thing does not mean you have experience in everything related to it. And even if you have experience in the right things, and it "works", it doesn't mean that things can't be improved.

I'm not expecting people to go "Oh wait, we really want you", rather I'm expecting the opposite "Well you were never any good anyway you self-important prick".

**At this rate, without a change of attitude / policy, unless SpaceManiac and/or his friends end up implementing it themselves, Glowstone will never, ever have a redstone framework.**

Unless my motivation comes back to me, I'm not doing more on this. Considering that the Keen Dreams source code has been recently released, I'll probably be focused on porting that.

lukecfairchild:
-_-

daboross:
The code you've written does do quite a bit though, and is an awesome base for redstone work. If I can find the time I'll see what I can do to implement redstone torch fizzling and create a new PR.

EntityReborn:
While your efforts were awesome, as we have explained before, there are certain ways to do things. Blaming _our_ attitude as being incorrect, while continuing with an attitude that itself is _not_ compatible with cooperating as a team, is quite frankly hypocritical. Your code is not really an issue here. For the most part, it's a great step. Manipulative events such as this, however, is downright childish. I previously had a lot of respect for you and what you can accomplish. Unfortunately, this is no longer true.

iamgreaser:
At the end of the day, I feel it's a political issue and somehow I'm incompatible. Perhaps someone more politically-aligned can push _something_ through. Who knows.

@daboross, go nuts. See what you can do.

SpaceManiac:
Hey, sorry you ran into trouble, and for not personally responding to your PR sooner. I appreciate the effort you gave so far and hope it ends up being useful for a redstone implementation in the long run. I'll look into making the contribution process more friendly, because in particular I agree with your opinion that expecting the PR to implement _all_ redstone in one go is pretty unreasonable, but I think a proper roadmap of in what order to implement bits of the redstone engine would be needed to proceed in a sane way. (Also, please understand that your PR is huge and I haven't reviewed it myself yet, so I can't speak for your code itself.)

dequis:
I'd like to highlight this part:

> While I accept that people use version control differently, there's no point in having exactly one branch that's kept as stable as possible when Git allows you to use several branches. Seriously, go add a "stable" branch and use "master" for the rough stuff. Or maybe add a "testing" and/or "unstable" branch. Different people have different requirements, and I'd be more than happy to use a bleeding-edge version even if it's full of bugs - aren't you glad PRs are a thing?

PR #225, opened 2014-09-08
Properly handle SIGTERM

lukasraska:
Initiate server shutdown on SIGTERM to save worlds and other data

### DIFF
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -87,6 +87,14 @@ public final class GlowServer implements Server {
             server.start();
             server.bind();
             logger.info("Ready for connections.");
+
+            Runtime.getRuntime().addShutdownHook(new Thread() {
+                @Override
+                public void run() {
+                    logger.warning("Caught SIGTERM, shutting down server");
+                    server.shutdown();
+                }
+            });
         } catch (Throwable t) {
             logger.log(Level.SEVERE, "Error during server startup.", t);
             System.exit(1);


SpaceManiac:
I think this might belong in ConsoleHandler's constructor rather than where you have it right now, thoughts?

SpaceManiac:
``` irc
<DarkKnightCZ> ok then... by the way, about #225, i don't think it should be in ConsoleManager, you can get sigterm from OS and if server is not properly inicialized, it shouldn't be tried to shut down and save everything, but i might be wrong as well :)
<SpaceManiac> DarkKnightCZ: something just feels inelegant about shoving it there
<DarkKnightCZ> yup
<SpaceManiac> I'm thinking maybe, at the end of GlowServer#start, call consoleManager.addShutdownHook() which does that
<SpaceManiac> alternatively, move the inner class into ConsoleManager (non-anonymous) and reference it there
<DarkKnightCZ> yup, it might be better
<SpaceManiac> mainly I want to avoid anonymous class in GlowServer
```

turt2live:
Sorry - bot fail :(

gdude2002:
I don't believe this is actually broken, but from time to time the build server gets the following..

```
From http://github.com/DarkKnightCZ/Glowstone
* [new branch]      bungee     -> PR/bungee
* [new branch]      entities   -> PR/entities
* [new branch]      gh-pages   -> PR/gh-pages
* [new branch]      iterable   -> PR/iterable
* [new branch]      master     -> PR/master
* [new branch]      mc-1.8     -> PR/mc-1.8
* [new branch]      misc       -> PR/misc
* [new branch]      rails      -> PR/rails
* [new branch]      scoreboard -> PR/scoreboard
error: Your local changes to the following files would be overwritten by merge:
        src/main/java/net/glowstone/GlowWorld.java
        src/main/java/net/glowstone/entity/GlowPlayer.java
Please, commit your changes or stash them before you can merge.
Aborting
```

This is really weird as there are no local changes. The script we're using to merge the branches is as follows..

``` sh
USERNAME="DarkKnightCZ"
BRANCH="misc"

echo "Adding remote.."
git remote add PR http://github.com/$USERNAME/Glowstone.git

if [ $? -ne 0 ]; then
    exit 1
fi

echo "Fetching remote.."
git fetch PR

if [ $? -ne 0 ]; then
    exit 1
fi

echo "Rebasing.."
git rebase PR/$BRANCH

if [ $? -ne 0 ]; then
    exit 1
fi

echo "Done!"
```

Any ideas?

Build: http://bamboo.gserv.me/browse/GLOW-SRV-PR225-95

turt2live:
@gdude2002 works fine for me, so I'm not sure what's going on... 

[Full build log](http://i.turt2live.com/yBqp.txt)

dequis:
> ```
>    src/main/java/net/glowstone/GlowWorld.java
>    src/main/java/net/glowstone/entity/GlowPlayer.java
> ```

Those are the two files which had their newlines completely changed in 74efe319c929a5bcf7af8392c115e8380be6caac and fixed in a357d2c62661ac89910688e499e849f026c9e7dc. Commits in the middle of those two confuse git.

Just add a `git reset --hard` and get rid of that noise, possibly coming from other PRs. I bet this is irrelevant for this PR - the only potential issues may be related to the fact that it's older than the newline changes

SpaceManiac:
I appreciate the contribution, but testing seems to indicate a shutdown hook is not an appropriate time to attempt the normal shutdown process. I may revisit this in the future since SIGTERM handling would certainly be useful.

PR #228, opened 2014-09-08
Display crafting table inventory as it's supposed to instead of a chest

### LINKED ISSUE #202: Wrong Inventory for Workbench, moves all items
When you open a workbench a default chest inventory (with a single line = 9 slots) shows up.
Additionally all items are moved one position rightwards while you have the table open (probably caused be the issue stated above). When you close the table the move to their original loaction.

Johni0702:
Minecraft seems to only show the crafting inventory correctly - regardless of inventory type -  when the slot size is set to 0.
This PR takes care of that and resolves #202


Johni0702:
This is going to be included in a future PR by @wow4201 containing further GUIs.

PR #207, opened 2014-09-07
Rogue threads are completely obliterated now

mastercoms:
Added better logic for thread sleeping, better rogue thread stopping, and finally, the shutdown thread stops the scheduler and any remaining tasks. Planning to add a graceful shutdown for scheduling later. (where it stops any new tasks from being created, then waits for any current tasks to be completed for a second or two, and then forcefully shuts down now)
Fixes #179

PR #231, opened 2014-09-08
Config option to turn off spawn chunk persistance

mastercoms:
This commit will give the option to turn off keeping spawn loaded in the config.
world:
  keep-spawn-chunks: false

That would not persist any spawn chunks. It is true by default, though.

EDIT: look at the new version: #233

### DIFF
--- a/src/main/java/net/glowstone/util/ShutdownMonitorThread.java
+++ b/src/main/java/net/glowstone/util/ShutdownMonitorThread.java
@@ -45,11 +45,11 @@ public class ShutdownMonitorThread extends Thread {
                 GlowServer.logger.warning("\tat " + trace.toString());
             }
 
-            //ask nicely to kill them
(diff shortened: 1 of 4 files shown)


src/main/java/net/glowstone/util/ShutdownMonitorThread.java
```diff
@@ -39,15 +40,16 @@ public void run() {
...
-            // really get it out of there
+            // ask nicely to kill them
             thread.interrupt();
-            thread.stop();
         }
+        // kill them forcefully
+        System.exit(1000);
     }
 
 }
```
ZephireNZ, line 55: Looks like this got accidentally pulled in from #207

mastercoms:
@ZephireNZ Yes, I closed this pull request because of that reason, it was in the wrong repo. 
Fixed pull is #233

PR #232, opened 2014-09-08
Add Checkstyle

HenrySlawniak:
Add a checkstyle config in order to more easily make sure that contributions meet guidlines.

Run using `gradle check`


src/main/java/net/glowstone/generator/populators/CavePopulator.java
```diff
@@ -67,7 +67,7 @@ public void run() {
                         }
                     }
                 }
-            }.start();
+            } .start();
```
turt2live: whitespace

turt2live:
Also you can just enable showing of hidden characters. does the same thing

src/main/java/net/glowstone/generator/populators/CavePopulator.java
SpaceManiac: The plugin complains about missing whitespace after a `}`, which is justifiable but could maybe be fixed in case this becomes a common construct at some point in the future.

config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version='1.0'?>
+<!DOCTYPE module PUBLIC '-//Puppy Crawl//DTD Check Configuration 1.2//EN' 'http://www.puppycrawl.com/dtds/configuration_1_2.dtd'>
+<module name='Checker'>
+  <module name='TreeWalker'>
+    <property name='tabWidth' value='4'/>
+    <module name='GenericWhitespace'/>
+    <module name='MethodName'>
+      <property name='format' value='^[a-z_][a-zA-Z0-9_]*$'/><!-- Cover _INVALID_ methodnames -->
```
SpaceManiac: Do the default options deny method names with `_` in them? If so, could you make this pattern accept _only_ `_INVALID_` methods, rather than underscores in general?

src/main/java/net/glowstone/generator/populators/CavePopulator.java
Ribesg: Should definitly be fixed. This whitespace is ugly as hell.

HenrySlawniak: It could be fixed in three ways
- Wrapping the Thread in parenthesis
- Letting checkstyle ignore this special case
- Assign the Thread a variable and start it on the next line

Thoughts?

src/main/java/net/glowstone/entity/meta/MetadataIndex.java
```diff
@@ -98,8 +98,7 @@
     ITEM_FRAME_ITEM(2, ITEM, ItemFrame.class),
     ITEM_FRAME_ROTATION(3, BYTE, ItemFrame.class),
 
-    ENDER_CRYSTAL_HEALTH(8, INT, EnderCrystal.class),
-    ;
+    ENDER_CRYSTAL_HEALTH(8, INT, EnderCrystal.class);
```
Ribesg: Doesn't seem like something we want, it's always better not to modify unrelated stuff when adding something.
Here adding something to the enum would modify  the `ENDER_CRYSTAL_HEALTH` line.

src/main/java/net/glowstone/generator/populators/CavePopulator.java
SpaceManiac: This construct is actually pretty rare and kind of ugly to exist on its own. Go with third option I think. This code isn't even really used.

HenrySlawniak:
Unless anything else arises, the only thing left is whether or not to add check to the default tasks.

HenrySlawniak:
Realized that it wasn't checking for tab characters.

dequis:
Trivial merge conflict in build.gradle (just add lines from both sides)

"PR: Formating Concerns" why?

turt2live:
@dequis Minor indentation issues

turt2live:
@meggawatts Your PR is currently not mergable. Please resolve the merge conflicts as soon as possible, thanks!

HenrySlawniak:
@turt2live Should be good to merge now ;D

SpaceManiac:
Thanks for the help! 62590b77810c8ba0c3fc1d02c5aa2f8bd2dcab50 is based on a combination of your config and that of SpongeAPI.

PR #233, opened 2014-09-08
Config option to turn off spawn chunk persistance

mastercoms:
This commit will give the option to turn off keeping spawn loaded in the config.
world:
keep-spawn-chunks: false

That would not persist any spawn chunks. It is true by default, though.

### DIFF
--- a/src/main/java/net/glowstone/util/ServerConfig.java
+++ b/src/main/java/net/glowstone/util/ServerConfig.java
@@ -307,6 +307,7 @@ public final class ServerConfig {
         GENERATOR_SETTINGS("world.gen-settings", "", Migrate.PROPS, "generator-settings"),
         ALLOW_NETHER("world.allow-nether", true, Migrate.PROPS, "allow-nether"),
         ALLOW_END("world.allow-end", true, Migrate.BUKKIT, "settings.allow-end"),
+        PERSIST_SPAWN("world.keep-spawn-chunks", true),
 
         // game props
         GAMEMODE("game.gamemode", "SURVIVAL", Migrate.PROPS, "gamemode"),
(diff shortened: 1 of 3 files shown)


SpaceManiac:
Is there an advantage to including this in core over letting a plugin do it? (Multiverse, for example, supports it).

src/main/java/net/glowstone/util/ServerConfig.java
```diff
@@ -307,6 +307,7 @@ private boolean migrate() {
         GENERATOR_SETTINGS("world.gen-settings", "", Migrate.PROPS, "generator-settings"),
         ALLOW_NETHER("world.allow-nether", true, Migrate.PROPS, "allow-nether"),
         ALLOW_END("world.allow-end", true, Migrate.BUKKIT, "settings.allow-end"),
+        PERSIST_SPAWN("world.keep-spawn-chunks", true),
```
turt2live: Maybe use "keep spawn loaded" or similar. This currently implies that the chunks will be kept across worlds (someone will ask it eventually)

mastercoms:
@SpaceManiac In future coding, we might need to know more information about spawn and it would be preferable to use definitive server code rather than plugin code.
Also, it allows server owners to tweak these settings without adding extra bloat like Multiverse would. Sometimes people want to configure the spawn persistence but don't want a new world manager.

src/main/java/net/glowstone/util/ServerConfig.java
mastercoms: @turt2live Will do.

turt2live:
@mastercoms This PR is currently not mergable. Please fix that ASAP!

turt2live:
@mastercoms still has merge conflicts :(

mastercoms:
@turt2live Ok, I'll have time over the weekend to fix it. Sorry for the delay!

turt2live:
@mastercoms Please update your PR as soon as possible.

mastercoms:
@turt2live I will make a new PR for it.

PR #130, opened 2014-08-22
Add blocks sounds

BeYkeRYkt:
https://github.com/SpaceManiac/Glowstone/issues/99

Issue #99, opened 2014-06-07
Block place sound effects are all wood


dequis:
Probably just need to send a sound effect packet? http://wiki.vg/Protocol#Sound_Effect

SpaceManiac:
Leaving this open for the fact that all blocks sound like wood right now and this should be fixed at some point.

dequis:
So... added "are all wood" to the title.

mastercoms:
Glowstone++ has custom block place sounds.

PR #242, opened 2014-09-09
Add support for variable block sounds

ghost:
This adds support for blocks to override the default placement sound with both variable volume and pitch. I created a new BlockSound class that encapsulates these values and can be overridden should a block need variable volume/pitch/sound etc. The pitch has a slight random deviation of +/- 0.25.


src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -51,6 +53,14 @@ protected final void setDrops(ItemStack... drops) {
             return Collections.unmodifiableList(drops);
         }
     }
+    
+    /**
+     * Gets the sound that will be played when a player places the block.
+     * @return The sound to be played
+     */
+    public BlockSound getSound() {
```
Tonodus: I'd prefer getPlacingSound() asit'd be clearer and maybe sometime there exists a getBreakSound() or getDiggSound() method, too.

src/main/java/net/glowstone/block/BlockSound.java
```diff
@@ -0,0 +1,33 @@
+package net.glowstone.block;
+
+import java.util.Random;
+
+import org.bukkit.Sound;
+import org.bukkit.block.Block;
+
+public class BlockSound {
```
Tonodus: Why BlockSound as class name? With little changes it can be used to play Songs to players / general locations. So sth. more generic would be nice (maybe GlowSound?)

ghost:
It appears that the placement and breaking sounds are the same and AFAIK the server does not handle playing the breaking sounds. I may be wrong though, so I put both methods in just in case.

src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -146,8 +164,7 @@ public final void rightClickBlock(GlowPlayer player, GlowBlock against, BlockFac
         newState.update(true);
 
         // play a sound effect
-        // todo: vary sound effect based on block type
-        target.getWorld().playSound(target.getLocation(), Sound.DIG_WOOD, 1, 1);
+        getPlaceSound().play(target.getLocation());
```
daboross, line 203: What if we added a `playSound(Location, GlowSound)` method to world instead of having a `.play(Location)` method on GlowSound?

EntityReborn, line 203: This actually follows more of an OOP style, currently. I personally prefer this.

src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -51,6 +53,22 @@ protected final void setDrops(ItemStack... drops) {
             return Collections.unmodifiableList(drops);
         }
     }
+    
+    /**
+     * Gets the sound that will be played when a player places the block.
+     * @return The sound to be played
+     */
+    public GlowSound getPlaceSound() {
+        return new GlowSound(Sound.DIG_WOOD); // default place sound
```
EntityReborn: Might be easier to maintain if for each BlockType, have a field specifying the sound, and reference that field here, instead of having to override sound methods like this.

turt2live: That would be ideal

src/main/java/net/glowstone/util/GlowSound.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Plays the sound at the given location with a slight variation in volume.
+     * @param location Location of the block producing this sound
+     */
+    public void play(Location location) {
+        // Sound pitch has a random deviation of +/- 0.25.
+        location.getWorld().playSound(location, sound, volume, pitch + random.nextFloat() * 0.5F - 0.25F);
+    }
+}
```
EntityReborn: Need a newline here.

src/main/java/net/glowstone/block/blocktype/BlockType.java
ghost, line 203: That could be useful. I believe the method in GlowSound should stay, though, since oftentimes we want to play a sound at a given location, and can just type `sound.play(location)` instead of having to get the world manually like `location.getWorld().playSound(location, sound)`. To play it to a single player there could be a `playToPlayer(Location, Player)`.

EntityReborn, line 203: Both methods would be useful, yes.

daboross, line 203: I guess if we were to have a world.playSound() method we would also have to add a class equivalent to GlowSound to Bukkit, so that the implementation doesn't leak through to the API. The current method works.

SpaceManiac:
A `GlowSound` already exists in `net.glowstone.constants`. Perhaps a name like `SoundInfo` would be more appropriate?

turt2live:
Ticket: #99 (I think?)

turt2live:
Sorry - bot fail :(

dequis:
What's the difference between this and #130?

turt2live:
@coasterman10 this doesn't implement the sounds like in #130 (thanks @dequis). Could that be added? Thanks!

(going to spam #130 as well. One solution will be pulled)

ghost:
This solution has support for variable volume and pitch (which is present for some blocks such as minecart tracks). Also is there any way to keep my fork in sync without messing up PR's with unnecessary commits?

turt2live:
You mean like the `Merge branch 'master' of ...` stuff?

ghost:
Yeah, that. I suppose I should make PR's from other branches like everyone else is?

daboross:
@coasterman10 Which command did you use to merge it?

turt2live:
Here's your series of commands (assuming 'upstream' points to this repo and not yours):

`git pull upstream master`
`git rebase origin/master`

But the `Merge branch...` stuff is fine as-is. It can be eliminated in post.

ghost:
Yeah I did `git pull upstream master` and that made a mess of a commit.

turt2live:
Yea, the rebase will fix it (I'm 90% sure at least). But the messy commit isn't a concern to us.

ghost:
Closing this for now due to inactivity and that I accidentally made it on my master branch. If anyone is still interested I can open a new PR from a different branch of mine.

SpaceManiac:
Thanks for the consideration; I've saved a copy of the diff locally and will try to review it when I can.

PR #246, opened 2014-09-10
Implement Item Frame placement.

wow4201:
Includes loading and saving of the item_frame.
This currently does not support items inside of item_frame.


gdude2002:
Why was this closed?

Issue #245, opened 2014-09-09
IndexOutOfBoundsException / UnpooledHeapByteBuf


tom-spalding:
Hi, server runs; however I get the following errors.

By client:

```
Connection Lost

Internal Exception: io.netty.handler.codec.DecoderException: java.lang.IndexOutOfBoundsException: readerIndex(1658190) + length(24832) exceeds writerIndex(1674062): UnpooledHeapByteBuf(ridx: 1658190, widx: 1674062, cap: 1674062/1674062)
```

By server:

```
~/minecraft# java -jar glowstone-0.0.1-SNAPSHOT.jar 18:42:39 [INFO] This server is running Glowstone version 1.8-16-gb6a27d4-b56 (Implementing API version 1.7.9-R0.1)
18:42:39 [INFO] Created default config: config/glowstone.yml
18:42:41 [INFO] Recipes: 259 shaped, 58 shapeless, 22 furnace, 15 fuels.
18:42:41 [INFO] Preparing spawn for world...
18:42:42 [INFO] Preparing spawn for world: 39%
18:42:43 [INFO] Preparing spawn for world: 83%
18:42:43 [INFO] Preparing spawn for world: done
18:42:43 [INFO] Preparing spawn for world_nether...
18:42:44 [INFO] Preparing spawn for world_nether: 58%
18:42:45 [INFO] Preparing spawn for world_nether: done
18:42:45 [INFO] Preparing spawn for world_the_end...
18:42:47 [INFO] Preparing spawn for world_the_end: 71%
18:42:47 [INFO] Preparing spawn for world_the_end: done
18:42:47 [INFO] Created default config: config/commands.yml
18:42:47 [INFO] Created default config: config/permissions.yml
18:42:47 [INFO] Binding to address: 0.0.0.0/0.0.0.0:25565...
18:42:47 [INFO] Ready for connections.
18:46:38 [INFO] starvapor [/130.212.130.56:54823] connected, UUID: d0e1ced8-ee23-4293-8c48-3920b6a49d60
18:46:38 [INFO] starvapor joined the game
18:46:38 [INFO] starvapor: ClientSettingsMessage{locale='en_US', viewDistance=12, chatFlags=0, chatColors=true, skinFlags=127}
18:46:39 [INFO] starvapor [/130.212.130.56:54823] disconnected: read error: java.io.IOException: Connection reset by peer
18:46:39 [INFO] starvapor left the game
>
```

[...]

Issue #136, opened 2014-09-04
Error while loading chunk on first run


SpaceManiac:
``` irc
<kookster> first run
<kookster> i opped myself
<kookster> then i set my gamemode to creative
<kookster> started flying
```

Presumably the chunk shouldn't have even been written yet, so it's odd to see an error loading it.

```
18:59:16 [INFO] This server is running Glowstone version 1.8-1-g40d9363-b46 (Imp
lementing API version 1.7.9-R0.1)
18:59:16 [INFO] Created default config: config\glowstone.yml
18:59:17 [INFO] Recipes: 259 shaped, 58 shapeless, 22 furnace, 15 fuels.
18:59:17 [INFO] Preparing spawn for world...
18:59:18 [INFO] Preparing spawn for world: 14%
18:59:19 [INFO] Preparing spawn for world: 33%
18:59:20 [INFO] Preparing spawn for world: 48%
18:59:21 [INFO] Preparing spawn for world: 66%
18:59:22 [INFO] Preparing spawn for world: 84%
18:59:23 [INFO] Preparing spawn for world: done
18:59:23 [INFO] Preparing spawn for world_nether...
18:59:24 [INFO] Preparing spawn for world_nether: 28%
18:59:25 [INFO] Preparing spawn for world_nether: 63%
18:59:26 [INFO] Preparing spawn for world_nether: 95%
18:59:26 [INFO] Preparing spawn for world_nether: done
18:59:26 [INFO] Preparing spawn for world_the_end...
18:59:27 [INFO] Preparing spawn for world_the_end: 43%
18:59:28 [INFO] Preparing spawn for world_the_end: 91%
18:59:28 [INFO] Preparing spawn for world_the_end: done
18:59:28 [INFO] Created default config: config\commands.yml
18:59:28 [INFO] Created default config: config\permissions.yml
18:59:28 [INFO] Binding to address: 0.0.0.0/0.0.0.0:25565...
18:59:29 [INFO] Ready for connections.
19:01:10 [INFO] [/192.168.1.119:55020] kicked: Outdated client!
19:02:26 [INFO] [/192.168.1.119:55082] kicked: Outdated client!
19:03:25 [INFO] kookster [/192.168.1.119:55142] connected, UUID: 273758c0-820b-4e75-bc72-92addaa1db2e
19:03:25 [INFO] kookster joined the game
[...]

Issue #127, opened 2014-08-17
CME in GlowPlayer#streamBlocks()


SpaceManiac:
```
14:39:02 [SEVERE] Error occurred while pulsing world world
java.util.ConcurrentModificationException
        at java.util.HashMap$HashIterator.nextEntry(HashMap.java:926)
        at java.util.HashMap$KeyIterator.next(HashMap.java:960)
        at java.util.AbstractCollection.addAll(AbstractCollection.java:341)
        at java.util.HashSet.<init>(HashSet.java:117)
        at net.glowstone.entity.GlowPlayer.streamBlocks(GlowPlayer.java:453)
        at net.glowstone.entity.GlowPlayer.pulse(GlowPlayer.java:359)
        at net.glowstone.GlowWorld.pulse(GlowWorld.java:362)
        at net.glowstone.scheduler.WorldScheduler$WorldThread.run(WorldScheduler.java:48)
```

turt2live:
@SpaceManiac reproduction information? (I doubt there's much :( )

PR #247, opened 2014-09-10
Fix disconnect when sending empty chunks

coelho:
When we send an empty chunk, we are expected to send the biome data. As of the current build of Glowstone, we do not do this, and it causes the client to disconnect. This PR fixes that issue.

### DIFF
--- a/src/main/java/net/glowstone/GlowChunk.java
+++ b/src/main/java/net/glowstone/GlowChunk.java
@@ -656,7 +656,7 @@ public final class GlowChunk implements Chunk {
             }
             byteSize += sectionCount * sectionSize;
         }
-        
+
         if (entireChunk) {
             byteSize += 256;  // + biomes
         }
--- a/src/main/java/net/glowstone/GlowChunk.java
+++ b/src/main/java/net/glowstone/GlowChunk.java
@@ -645,53 +645,53 @@ public final class GlowChunk implements Chunk {
             }
         }
 
-        // break out early if there's nothing to send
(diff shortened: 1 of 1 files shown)


turt2live:
What disconnect?

coelho:
If you are in a world with empty chunks (ie. the end) then you will be disconnected due to a Netty DecoderException as it tries to read the biomes. #245 is related, but it was incorrectly closed.

turt2live:
Does this touch or even look at #136 or #127 ? #245 was closed. If no ticket currently exists, can you create one please?

SpaceManiac:
Dabo on IRC independently confirmed the problem's existence and that it is fixed by this PR.

daboross:
I reproduced this when testing SkyWars (download http://ci.daboross.net/d/SkyWarsParent_SkyWars_MainBuild/SkyWars.jar). The plugin uses a generator which returns empty chunks, then populates with a few islands where the players stand.

A screenshot of the error I was getting can be found at http://qx.lc/wQQ9. The error no longer exists with this PR.

turt2live:
Looks as though it's working. Looks like a solid implementation as well. Ready for pulling :)

SpaceManiac:
Merged in 3641cf44af60cd504c32aabd2daef6ea9ca7bfd7.

xvento:
image

I have this problem please help

turt2live:
@xvento Please open a new ticket with more information or consult our forums. Thanks!

PR #250, opened 2014-09-10
Add bed placement

vurpo:
This makes bed placement possible. Beds get placed with the foot block on the clicked block, and the head block one block away from the foot block in the opposite direction from the player, just like in vanilla Minecraft. Placement with the head block in the air or inside another block is not possible.


src/main/java/net/glowstone/block/blocktype/BlockBed.java
```diff
@@ -0,0 +1,69 @@
...
+public class BlockBed extends BlockType {
+
+    public BlockBed() {
+        setDrops(new ItemStack(Material.BED));
+    }
+
+    @Override
+    public void placeBlock(GlowPlayer player, GlowBlockState state, BlockFace face, ItemStack holding, Vector clickedLoc) {
+        BlockFace direction;
+        double rot = player.getLocation().getYaw() % 360;
```
EntityReborn: Stuff like this should probably be done in `getYaw()` to avoid duplicated/buggy code.

turt2live: Or add a method to `Location`: `getRotation` and document it returns degrees.

turt2live:
Ticket: #119

turt2live:
Looks good. Besides the obvious physics problems (out of scope), this seems ready to be pulled.

http://i.turt2live.com/WAlA.jpg

src/main/java/net/glowstone/block/blocktype/BlockBed.java
```diff
@@ -0,0 +1,69 @@
...
+            } else {
+                GlowServer.logger.warning("Placing " + getMaterial().name() + ": MaterialData was of wrong type (" + data.getClass().getName() + ")");
+            }
+        }
+    }
+
+    @Override
+    public void afterPlace(GlowPlayer player, GlowBlock block, ItemStack holding) {
+        if (block.getType() == Material.BED_BLOCK) {
+            BlockFace direction;
```
turt2live: all of these variables shouldn't be declared and then set. Instead they should be set when declared.

turt2live:
@vurp0 correction to previous comment. Please see diff.

vurpo:
@turt2live Fixed that now.

turt2live:
Sorry - bot fail :(

turt2live:
Appears to work, thanks!

SpaceManiac:
Merged in 7dae0f0ff64146b54821e417bbf2024d221d299b.

PR #264, opened 2014-09-12
Implement BlockContainer.getDrops()

dequis:
Drops the container block itself, and all the items from its inventory.

Fixes the weird looking item that chests dropped (was chest + data)

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockContainer.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockContainer.java
@@ -5,7 +5,10 @@ import net.glowstone.block.entity.TEContainer;
 import net.glowstone.block.entity.TileEntity;
 import net.glowstone.entity.GlowPlayer;
 import org.bukkit.block.BlockFace;
+import org.bukkit.inventory.ItemStack;
 import org.bukkit.util.Vector;
+import java.util.Collection;
+import java.util.LinkedList;
 
 /**
  * Base BlockType for containers.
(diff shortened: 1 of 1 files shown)


dequis:
Screenshot of the weird-looking item:



This happened because the default implementation of BlockType.getDrops does the following:

```
return Arrays.asList(new ItemStack(block.getType(), 1, block.getData()));
```

My implementation removes the getData to return a pure chest block:

```
list.add(new ItemStack(block.getType(), 1));
```

And additionally includes the whole inventory as part of that list, so items are not lost.

JeromSar:
Looks really good @dequis. You should probably warn when the expected block is not a `TEContainer`. :)

dequis:
Is there any case in which the tile entity of the expected block is not a `TEContainer`?

As far as I know, since this is `BlockContainer`, its subclasses (such as `BlockChest`) create tile entities that subclass from `TEContainer` (such as `TEChest`)

JeromSar:
@dequis Correct, but you can never be too sure. We do the same for `BlockType` classes. https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/block/blocktype/BlockDispenser.java#L34

turt2live:
Sorry - bot fail :(

turt2live:
Looks good!

SpaceManiac:
Merged in d70a3e8133a94e1276f9f6478415327cc2e2a561.

dequis:
:heart:

PR #291, opened 2014-09-16
Redesign gh-pages branch

HenrySlawniak:
Just a super simple bootstrap and bootswatch site, live demo here: http://meggawatts.github.io/Glowstone/


Aaron1011:
@meggawatts: This looks **really** nice!

Postremus:
Little issue: If you are on the Downloads page and want to go to the
Api-Docs, you get a 404

2014-09-16 23:15 GMT+02:00 Aaron1011 notifications@github.com:

ZephireNZ:
Your {{site.baseurl}} isn't working properly. Apparently you need to set the baseurl in the config file (http://jekyllrb.com/docs/github-pages/#project-page-url-structure)

stuntguy3000:
Why did you remove  **target="_blank"**?

HenrySlawniak:
atom.io did some weird things, fixed now

SpaceManiac:
Looks pretty, I like it. It seems odd to me, though, to include the javadocs (which is compiled output) in the source control. Maybe link to greatman's javadoc hosting at http://build.greatmancode.com/job/Glowkit/javadoc/ instead?

HenrySlawniak:
Didn't realize there were already javadocs for Glowkit somewhere. Diff is a lot more manageable now too. :D

HenrySlawniak:
I feel like the three little blurbs could be simplified somehow. I would love to hear anyone's thoughts.

SpaceManiac:
Maybe something along the lines of...

> Glowstone's code is completely original and licensed under the permissive MIT license. There is no reliance on decompiled code.
> 
> Simplicity breeds performance. Glowstone is built with both of these in mind, and offers benefits in situations where vanilla game features aren't needed.

Not final. I like the one about Bukkit.

HenrySlawniak:
Been playing around with adding a little bit of color, that combined with the simplified blurbs seems to really make a difference.

Nickforall:
That looks much better :+1: , but I like the leaf more for lightweight instead of the lightning.

SpaceManiac:
A few comments:
- I don't think the "Minecraft" in the main header needs to be a link. There's a link later and IMO it looks nicer when the text style is uniform.
- The main header feels like it's got slightly too much vertical whitespace around it but this might just be me. Other parts of the page have this but not as bad.
- I know you just brought this from the old site but I don't think a Travis CI link is actually necessary. There's nothing there that's really useful to someone using Glowstone. Maybe link to the Github wiki instead?
- The downloads page should maybe have some intro text before the table so it doesn't feel as cramped, not sure what though.
- I'm not sure what's up with `... Lightstone. As well as Lightstone"s contributors.`

Other than that I really like it so far. Great work!

SpaceManiac:
Another suggestion for the Downloads page: specify fixed column widths so that the headers don't rearrange themselves when the data is finished loading.

HenrySlawniak:
Not entirely sure what merge conflicts there are. Any git magicians able to figure it out?

SpaceManiac:
Probably just Github being annoyed about the fact it's `gh-pages` or something. If you consider this ready to merge, I'll take care of it.

HenrySlawniak:
I'm happy with it, so if you're happy with it, go for it.

SpaceManiac:
Merged in 52d79df143523fd60d08a56c4b05de17ce4a4a2d. Thanks!

PR #334, opened 2014-09-22
Move code for block rotation out of BlockSign

ZephireNZ:
This generifies the rotation code in BlockSign, as it is also used for banners and skulls.

Three changes:
- Changes BlockSign to use the Bukkit API to set its rotation, rather than setting the raw data directly.
- Add method to GlowEntity to get its full, intercardinal (eg north north east) direction in BlockFace.
- Add method to Position Util to convert between BlockFace and byte data value (common between rotatables)


ZephireNZ:
Ok, so apparently Minecraft uses two different starting points: South for Signs/Banners, and North for Skulls -_-
Also, the code right now says the player is facing in the opposite direction. Going to fix that, as it's counter intuitive.

SpaceManiac:
Closing in favor of #360, which includes these changes.

PR #336, opened 2014-09-22
Handle invalid arguments with a warning only

bendem:
This commit changes the server startup code so that it doesn't fail if it encounters an invalid argument.
This change is motivated by the fact that most server hosts provide the nogui option in their startup script and would thus prevent the glowstone servers from starting.


SpaceManiac:
Merged in d986e8e3f0c3094d007acf8ae101629bb678a7de.

PR #337, opened 2014-09-22
Typo

bendem:


dequis:
"wgen" probably means "world gen", which makes sense in this context. Sorry.

bendem:
You're totally right.

PR #345, opened 2014-09-25
Moving scripts to the bottom of page

choyiny:
Moving the scripts to the bottom of page to increase web performance.


dequis:
PROTIP: test your code before submitting a PR

(You're moving the jquery script tag after the code that uses `$`)

_includes/head.html
```diff
@@ -2,8 +2,6 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <script src="//ajax.googleapis.com/ajax/libs/jquery/1.10.2/jquery.min.js"></script>
-    <script src="//netdna.bootstrapcdn.com/bootstrap/3.2.0/js/bootstrap.min.js"></script>
     <link href="//netdna.bootstrapcdn.com/bootstrap/3.2.0/css/bootstrap.min.css" rel="stylesheet"/>
```
mastercoms: It would be good to remove this line, the bootswatch yeti theme includes Bootstrap CSS anyway.

_includes/foot.html
```diff
@@ -12,9 +12,12 @@
     </p>
   </div>
   </body>
+  <script src="//ajax.googleapis.com/ajax/libs/jquery/1.10.2/jquery.min.js"></script>
```
mastercoms, line 15: Perhaps change src to //code.jquery.com/jquery-1.11.0.min.js (this uses maxcdn, which also powers bootstrapcdn)

mastercoms:
Some older browsers don't support the //. Perhaps add https:// instead?

mastercoms:
@dequis This pull could remove scripting from the includes and just include it in the pages that need it:
<script src="jquery.js">
<script>
jquery code
</script>
</body>
</html>

and yes, it is recommended to add <script> tags in body.

SpaceManiac:
> Some older browsers don't support the //

What browsers?

ZephireNZ:
From StackOverflow (god bless):

> This behavior was part of RFC 1808 (Section 4) which is about 16 years old, so every major browser should (and does) support this.
> 
> Sadly, there's a bug with IE7 and -8 that will make them download the resources twice if a protocol-relative URL is used on a link or @import - which shouldn't be a big problem, but is ugly and should be kept in mind.

SpaceManiac:
Since all the URLs are from CDNs that can definitely handle this tiny additional load from faulty IE versions, I kind of want to keep that as it was before, in case I start serving glowstone.net over HTTPS in the future.

choyiny:
Putting js files at the bottom makes the website loading the mobile view instead of the web view (dont know why but the mobile version does not load on my phone @ glowstone.net but is able to be loaded in my fix)

i will fix to remove http again

SpaceManiac:
Merged in 181284f.

PR #346, opened 2014-09-25
Add Eclipse and IDEA plugins to build.gradle

jimmikaelkael:
This will provide a few tasks to gradle for building project files for both IDE.
Those who wish to import an Eclipse project will do:
./gradlew eclipse
Those who wish to import an IDEA project will do:
./gradlew idea

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,6 +8,8 @@ apply plugin: 'project-report' // Adds project report tasks
 apply plugin: 'com.github.johnrengelman.shadow' // Provides dependency shading
 apply plugin: 'net.glowstone.remapper' // Provides method remapping
 apply plugin: 'checkstyle' // Checks code style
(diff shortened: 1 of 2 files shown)


jimmikaelkael:
I added a 2nd commit as it's required that the project name match directory name so that it can be imported directly as a gradle project into Eclipse.

Let me know if you want me to squash this 2nd commit.

SpaceManiac:
Merged in ed97e6810dee31d32b925522a8e0fcb03f1d1d3b.

SpaceManiac:
Turns out that second commit is trouble - it changes all the file paths from 'glowstone' to 'Glowstone' as well. I think I'd prefer to see a workaround for loading things in Eclipse or a change which keeps the artifact names lowercase.

jimmikaelkael:
I'm sorry for that, an alternate way to do this is to simply set an eclipse project name.

See PR #352

PR #351, opened 2014-09-26
Double chests

### LINKED ISSUE #263: Double chests behave like two single chests
Click one side, get one chest with three rows of items.

Click the other side, get another chest with three rows of other items.

### LINKED ISSUE #329: Weird chest ghost block when making it with shift click.
Making a double chest by shift clicking the original chest sometimes does not work, it will create a ghost block, and an invisible chest block.

https://www.youtube.com/watch?v=f6e4xcEp4L8&feature=youtu.be

Tonodus:
This adds all the functionality for double chests:
- correct facing
- correct inventory (loading/saving)

The problem that currently exists is that Bukkit doesn't have MaterialData for TRAPPED_CHESTS. I added TRAPPED_CHEST as BlockChest too, since I assume that TRAPPED_CHESTS should also have a MaterialData of org.bukkit.material.Chest. If this is added to Glowkit it should automatically work with trapped chests correct, but since then placing trapped_chests gives you a warning in the console.
(See: [GlowkitPR](https://github.com/GlowstoneMC/Glowkit/pull/23)

Fixes #263 
Fixes #329

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockChest.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockChest.java
@@ -1,21 +1,59 @@
 package net.glowstone.block.blocktype;
 
 import net.glowstone.GlowChunk;
+import net.glowstone.GlowServer;
+import net.glowstone.block.GlowBlock;
 import net.glowstone.block.GlowBlockState;
 import net.glowstone.block.entity.TEChest;
 import net.glowstone.block.entity.TileEntity;
 import net.glowstone.entity.GlowPlayer;
+import org.bukkit.Material;
 import org.bukkit.block.BlockFace;
+import org.bukkit.block.BlockState;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.material.Chest;
 import org.bukkit.material.MaterialData;
 import org.bukkit.util.Vector;
 
+import java.util.ArrayList;
+import java.util.Collection;
+
 public class BlockChest extends BlockContainer {
+
+    private final boolean isTrapped;
+
+    public BlockChest(boolean isTrapped) {
+        this.isTrapped = isTrapped;
+    }
 
     @Override
     public TileEntity createTileEntity(GlowChunk chunk, int cx, int cy, int cz) {
         return new TEChest(chunk.getBlock(cx, cy, cz));
+    }
+
+    @Override
+    public boolean blockInteract(GlowPlayer player, GlowBlock block, BlockFace face, Vector clickedLoc) {
+        BlockState state = block.getState();
+        if (state instanceof org.bukkit.block.Chest) {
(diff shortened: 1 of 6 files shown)


src/main/java/net/glowstone/block/blocktype/BlockChest.java
```diff
@@ -1,33 +1,157 @@
...
+
+            BlockFace normalFacing = getOppositeBlockFace(player.getLocation(), false);
+
+            Collection<BlockFace> attachedChests = searchChests(chestBlock);
+            if (attachedChests.isEmpty()) {
+                chest.setFacingDirection(normalFacing);
+                state.setData(chest);
+                return;
+            } else if (attachedChests.size() > 1) {
+                GlowServer.logger.warning("Chest placed near two other chests!");
```
ZephireNZ, line 76: Should we be logging for something like this in production?

src/main/java/net/glowstone/block/entity/TEContainer.java
```diff
@@ -25,6 +26,10 @@ public Inventory getInventory() {
         return inventory;
     }
 
+    public void setContent(ItemStack[] content) {
```
ZephireNZ: setContent<b>s</b> ?

src/main/java/net/glowstone/block/blocktype/BlockChest.java
Tonodus, line 76: Yes, because this should never happen: this case should be already handled by canPlaceAt(). But it can be removed of course, if wanted.

ZephireNZ, line 76: Ah, I missed that. Keep it :)

src/main/java/net/glowstone/inventory/GlowInventory.java
```diff
@@ -291,6 +297,22 @@ public void setContents(ItemStack[] items) {
         System.arraycopy(items, 0, slots, 0, items.length);
     }
 
+    /**
+     * Replace the current part of this content with the given one, starting by beginIndex.
+     *
+     * @param beginIndex The index to begin to replace the contents
+     * @param items The ItemStack that will replace the existing ones
+     */
+    public void replaceContent(int beginIndex, ItemStack... items) {
```
JeromSar: Should probably be `replaceContents`

src/main/java/net/glowstone/block/blocktype/BlockChest.java
daboross, line 76: Might want to comment explaining it for future readers though

turt2live:
Note to self/other reviewers: This must be reviewed in conjunction with #358

turt2live:
Closing in favour of #358. Thank you for your contribution!

PR #352, opened 2014-09-27
Add Eclipse project name to build.gradle

jimmikaelkael:
This is another way to fix Eclipse project name as it's required that the project name match directory name so that it can be imported directly as a gradle project into Eclipse.

As it has been discussed in PR #346, we want to preserve artifact names.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -117,6 +117,13 @@ checkstyle {
     configFile = file('etc/checkstyle.xml')
 }
 
+// Eclipse project name
+eclipse {
+    project {
+        name = 'Glowstone'
+    }
+}
+
 // Jar manifest information
 jar.manifest.mainAttributes(
         'Main-Class': mainClassName,


build.gradle
```diff
@@ -117,6 +117,13 @@ checkstyle {
     configFile = file('etc/checkstyle.xml')
 }
 
+// Eclipse project name
+eclipse {
+    project {
+        name = 'Glowstone'
```
JeromSar, line 123: As per https://github.com/GlowstoneMC/Glowstone/commit/1fa24a7bf6c2fc555217aa8e79dd3f74647170d3, this should be in lowercase.

bendem, line 123: > This is **another way** to fix Eclipse project name

As in, since the commit you linked causes problem, this is another try to resolve Eclipse's problem.

jimmikaelkael:
Nope, it's not gradle project name...
This commit is necessary otherwise you're unable to import the gradle project with Eclipse's gradle plugin.
This only changes Eclipse project files.

EDIT: @bendem was faster...

build.gradle
mastercoms, line 123: I also feel like glowstone should be lowercase, it unifies the naming across development, and you don't get any more features by making it capitalized.

jimmikaelkael, line 123: Maybe... but if you're cloning the repo without forcing a directory name, it's 'Glowstone' as per the github project name.

JeromSar, line 123: @jimmikaelkael For consistency, I think lowercase should be used here.

SpaceManiac:
Merged in 45039a25c41ef88779f1e365ea31fe74338fa4c6.

Issue #349, opened 2014-09-26
Sword prevents dig even in survival


SpaceManiac:
The sword dig checks in `DiggingHandler` need to be changed to operate only in creative mode.

mastercoms:
https://github.com/GlowstoneMC/Glowstone/blob/2ea5576273e89106a2e22ce3ba90c9ce8fc21e31/src/main/java/net/glowstone/net/handler/play/player/DiggingHandler.java#L52
needs to be changed to
if (player.getGameMode() == GameMode.CREATIVE && holding != null && EnchantmentTarget.WEAPON.includes(holding.getType())) {

ZephireNZ:
Fix on PR #356

PR #356, opened 2014-09-27
Check for creative and sword on break (fixes #349)

mastercoms:
Add a check for creative when canceling block breaking if the player has a sword in his hand.


SpaceManiac:
Merged in 2690d3f81459d46e58e1df661828a89070f41bf3.

PR #357, opened 2014-09-27
Renamed net package to network

ghost:
For more intuitively understanding

### DIFF
--- a/src/main/java/net/glowstone/ChunkManager.java
+++ b/src/main/java/net/glowstone/ChunkManager.java
@@ -1,5 +1,6 @@
 package net.glowstone;
 
+import net.glowstone.constants.GlowBiome;
 import net.glowstone.io.ChunkIoService;
 import org.bukkit.block.Biome;
 import org.bukkit.event.world.ChunkLoadEvent;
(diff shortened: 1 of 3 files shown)


mastercoms:
Why? this doesn't do anything what so ever to further glowstone, and might break things.

Postremus:
I think it is inuitive enoug as it is, because java has a "java.net"
package, too.

2014-09-27 23:21 GMT+02:00 Vladislavs Golubs notifications@github.com:

Tonodus:
Git doesn't handle file/folder renaming. As such this PR causes much trouble with the history of the related files, PR/forks working on this files etc. It isn't worth it IMO.

daboross:
'net' doesn't really mean much else besides network. In this context, I don't really see how it is any less intuitive. Plus, changing the package breaks any external stuff using Glowstone's internal networking, and messes with a ton of file history diffs.

ghost:
'net.glowstone.net' sounds weird

Wild1145:
It might just be me, but I see no file changes and infinite commits...

Tonodus:
@Wild1145 Seems like he deleted that commits.

mastercoms:
@ensirius Could you close this pull request if there is nothing to merge?

PR #371, opened 2014-09-29
Spawn the player next to the bed instead of inside it

GermanCoding:
This is my first pull request, so please say me if I did something wrong.

I made a little change in the respawn() function. The old code respawned the player in his bed. The new code should find a place next to the bed where the player could stand (2 air blocks). If there is no good place the player will respawn at the world's spawn.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,8 +15,3 @@ target
 !.gitmodules
 !.mailmap
 !.gitattributes
-*.json
-*.yml
-*.txt
-*.dat
-*.mca
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,8 @@ target
 !.gitmodules
 !.mailmap
 !.gitattributes
+*.json
+*.yml
+*.txt
+*.dat
+*.mca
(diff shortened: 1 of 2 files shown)


Tonodus:
You'd better use an array with all BlockFaces and a loop. Also you should take into consideration that a bed has a width of 2 blocks, so you havr to check other locations than the direct N/E/S/W relatives.

jimmikaelkael:
You have some style problems, it's the reason it doesn't build.

```
[ant:checkstyle] /home/travis/build/GlowstoneMC/Glowstone/src/main/java/net/glowstone/entity/GlowPlayer.java:631:17: '}' should be on the same line.
[ant:checkstyle] /home/travis/build/GlowstoneMC/Glowstone/src/main/java/net/glowstone/entity/GlowPlayer.java:632:24: WhitespaceAround: 'if' is not followed by whitespace.
[ant:checkstyle] /home/travis/build/GlowstoneMC/Glowstone/src/main/java/net/glowstone/entity/GlowPlayer.java:634:17: '}' should be on the same line.
[ant:checkstyle] /home/travis/build/GlowstoneMC/Glowstone/src/main/java/net/glowstone/entity/GlowPlayer.java:635:24: WhitespaceAround: 'if' is not followed by whitespace.
[ant:checkstyle] /home/travis/build/GlowstoneMC/Glowstone/src/main/java/net/glowstone/entity/GlowPlayer.java:635:94: WhitespaceAround: '{' is not preceded with whitespace.
[ant:checkstyle] /home/travis/build/GlowstoneMC/Glowstone/src/main/java/net/glowstone/entity/GlowPlayer.java:639:19: WhitespaceAround: 'if' is not followed by whitespace.
```

Issue #296, opened 2014-09-18
PlayerLoginEvent: event.getPlayer().isOnline() always returns true, even if the player hasn't logged in


dequis:
Test case plugin: [jar](https://github.com/dequis/bukkit-plugin-test-cases/releases/download/PlayerLoginEvent-isOnline/Test-PlayerLoginEvent-isOnline.jar), [source](https://github.com/dequis/bukkit-plugin-test-cases/commit/PlayerLoginEvent-isOnline)

Log in to test, see console for output.

Expected behavior (craftbukkit)

```
[07:48:28 INFO]: [Test] onPlayerLogin: player.isOnline() == false
```

Actual behavior:

```
07:50:52 [INFO] [Test] onPlayerLogin: player.isOnline() == true
```

AuthMe uses this return value to determine if the player is trying to log in twice:

https://github.com/Xephi/AuthMeReloaded/blob/master/src/main/java/fr/xephi/authme/listener/AuthMePlayerListener.java#L424

So the second attempt to login should return true.

dequis:
Possibly related:

#153 (prevents double login, same thing AuthMe does...)

#156 (a quit message is shown even after login failed)

Issue #156, opened 2014-09-05
Players blocked from a whitelisted server cause `XYZ has left the game` to be shown in chat.


basicer:

basicer:
![Example](https://www.screenshottr.us/v/afe110f57564bc97d380bca18d44654bc8b3ea42/a7376376f4316358fec04c846851cd91.png)

nevercast:
The commit you have there should check whitelistEnabled, otherwise there will never be any quit messages when the whitelist is empty and whitelist is disabled

PR #373, opened 2014-09-30
Fixed Player.isOnline() returning true when the player isn't already log...

jimmikaelkael:
...ged in

This is an attempt to fix issues #296 and #156. I've tested it and it fixed both issues, but this needs to be carefully reviewed to avoid unwanted side effects.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -245,6 +245,11 @@ public final class GlowPlayer extends GlowHumanEntity implements Player {
     private float walkSpeed = 0.2f;
 
     /**
+     * The player's online state.
+     */
+    private boolean online = false;
+
+    /**
      * Creates a new player and adds it to the world.
      * @param session The player's session.
      * @param profile The player's profile with name and UUID information.
@@ -730,7 +735,14 @@ public final class GlowPlayer extends GlowHumanEntity implements Player {
 
     @Override
     public boolean isOnline() {
-        return session.isActive();
+        return online;
+    }
+
+    /**
+     * Sets this player's online state.
+     */
+    public void setOnline() {
+        this.online = true;
     }
 
     @Override
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -735,7 +735,7 @@ public final class GlowPlayer extends GlowHumanEntity implements Player {
 
     @Override
     public boolean isOnline() {
-        return online;
+        return online && session.isActive();
     }
 
     /**
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -245,11 +245,6 @@ public final class GlowPlayer extends GlowHumanEntity implements Player {
     private float walkSpeed = 0.2f;
 
     /**
-     * The player's online state.
-     */
-    private boolean online = false;
-
-    /**
      * Creates a new player and adds it to the world.
      * @param session The player's session.
      * @param profile The player's profile with name and UUID information.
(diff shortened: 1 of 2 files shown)


turt2live:
I'm not sure if passing a state variable that is public is a good idea. This could be used by bad plugins to set the player as offline, therefore breaking many things. Is it possible to make this less exposed?

Otherwise, it works!

jimmikaelkael:
Yeah I will do those changes ASAP.

jimmikaelkael:
@turt2live rebased with the changes discussed above.

jimmikaelkael:
I added a commit to keep the GlowSession alive check in GlowPlayer.getOnline() to ensure plugins gets a valid check if they hold a reference on the player.

turt2live:
The change still allows for potentially bad plugins to incorrectly set the player as online. If possible, try to use similar logic as before where the networking portion is used (store the flag there?)

dequis:
Do we really care about plugins that intentionally misbehave that badly? It's not like this is part of the bukkit api...

It would be like the equivalent of a plugin going into NMS code and setting a player as online or offline. For what purpose, I don't know, but there's no way that can be done accidentally.

turt2live:
A plugin is just an example. It's still unsafe code and it doesn't belong where it currently is.

jimmikaelkael:
Commited another approach. The online state is in the session so it's less exposed.

turt2live:
Unintentional side effects of this PR:
- Players are not correctly disconnected for being banned/not whitelisted
  - Some are getting "reset by peer" errors
  - Some get the wrong message (I got "session servers under maintenance" once)
  - Someone got "you are not whitelisted" when the whitelist was on and they were banned once
- Players do not receive the correct message when disconnected
- Players can see parts of the world when they connect and are not whitelisted or are banned
  - May be out of scope (issue related to the message being shown previously)

The working part: `isOnline` is functional and the messages don't show \o/

jimmikaelkael:
@turt2live Seriously, have you looked at the changes ?
Have you tested that the same doesn't happen without this PR changes ?

I got all of the above on a clean master...

turt2live:
THe point of this PR, as described in the description, is to fix many of the listed issues. The issues were double checked against master with only the issues relating to this PR being listed.

jimmikaelkael:
Tested again with clean master:
- turn on whitelist and try to connect while not being whitelisted: I got unconsistent error messages like those you mentionned above.
- turn off whitelist and try to connect, ban me from console: same as above. Even worst if you use a shortcut as name: it kicks you for being ban (while not really applying it on the realname) while name shortcuts should never be used for bans. That may explain the "you are not whitelisted" when the whitelist was on and they were banned once.
- In the 2 attempts above I'm seing parts of the world before to be disconnected... with a wrong  message.

I'm getting too a strange exception sometimes:

```
09:37:49 [GRAVE] Error in network input
io.netty.handler.codec.DecoderException: java.lang.IndexOutOfBoundsException: readerIndex(3) + length(106) exceeds writerIndex(15): UnpooledUnsafeDirectByteBuf(ridx: 3, widx: 15, cap: 15)
```

jimmikaelkael:
I can't find any relationship with the problems you mentionned and this PR, so I'm closing it.
EDIT:
@turt2live the real source of the problem is that the JoinGame packet is sent in the GlowPlayer constructor: this shouldn't be the case unless we are sure the player can join...

dequis:
Dammit turt2live I told you not to expand the scope of PRs

turt2live:
I didn't get any of those issues on my version of a master O_o. 

If you're willing @jimmikaelkael, I'll test again and reopen this PR. I probably just screwed something up with copying jar files.

jimmikaelkael:
@turtlive that's not like I took it as being offensed, really.
But to me most of the problems metionned above exists already in master branch, and I'm almost 100% sure that the problem with wrong messages and player seeing the world briefly before to be kick is due to the JoinGame packet sent in GlowPlayer constructor: it's should be deffered after we can ensure the player can join.

I closed this issue because I don't feel I'm able to fix it all in once.

turt2live:
I'm not asking you to fix them all at once. I originally noticed discrepancies between "master" and this is all. If master does indeed have the same issues, then this is valid as is.

jimmikaelkael:
Well I'll try to make some tests to see if I can come up with something solid, then I'll reopen this PR if it's the case.

PR #381, opened 2014-10-03
[WIP] Continuation of redstone support

kaenganxt:
This will be a continuation of iamgreaser's redstone framework pull request (https://github.com/GlowstoneMC/Glowstone/pull/215), which was closed by him. I think its a very good start for redstone support in Glowstone, so i wanted to continue it.
## PRs:

As this pull request gets splitted. Here are references to the other PRs:
#436 - Basics
#439 - Button & Lever
## TODO:
- Get torches to burn out when flickering
- Pressure plates
- Daylight sensor
- Redstone comparator (comparing & subtracting; block state measuring maybe part of another PR)

I think redstone devices like dropper, dispenser and so on should be part of another PR.

I think it is ok to use his commits, since he said it to someone else, that it would be okay.
If not @iamgreaser, please just say it.

### DIFF
--- a/src/main/java/net/glowstone/RSManager.java
+++ b/src/main/java/net/glowstone/RSManager.java
@@ -1,0 +1,301 @@
+package net.glowstone;
+
+import net.glowstone.GlowServer;
+import net.glowstone.GlowWorld;
+import java.util.*;
+
+/**
+ * A class to manage redstone logic for a world.
+ * @author Ben Russell
+ */
+public class RSManager
+{
+    /**
+     * A manual enumeration of block types.
+     * TODO: Something better, perhaps?
+     */
+    public final int B_AIR = 0;
+    public final int B_SOLID = 1;
+    public final int B_RWIRE = 2;
+    public final int B_RTORCH = 3;
+    public final int B_RDIODE = 4; // Not supported yet
+
+    /**
+     * An immutable class for storing 3D world positions.
+     * TODO: Consider using Location (assuming it's hashable to the nearest int^3)
+     */
+    public final class RSPos
+    {
+        public final int x, y, z;
+
+        public RSPos(int x, int y, int z)
+        {
+            this.x = x;
+            this.y = y;
+            this.z = z;
+        }
+
+        @Override
+        public int hashCode()
+        {
+            int hash = 1;
+
+            hash = hash * 31 + this.y;
+            hash = hash * 53 + this.x;
+            hash = hash * 19 + this.z;
+
+            return hash;
+        }
+
+        @Override
+        public boolean equals(Object o)
+        {
+            if(o instanceof RSPos)
+            {
+                RSPos op = (RSPos)o;
+
+                return true
+                    && op.x == this.x
+                    && op.y == this.y
+                    && op.z == this.z;
+
+            }
+
+            return false;
+        }
+
+        @Override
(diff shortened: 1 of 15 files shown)


src/main/java/net/glowstone/RSManager.java
```diff
@@ -0,0 +1,526 @@
...
+     * A flag used to enable testing redstone by spamming the console.
+     * Best to only enable this if you are a developer.
+     */
+    public static final boolean DEBUG_REDSTONE = false;
+
+    /**
+     * An immutable class for storing 3D world positions.
+     */
+    //TODO: Where should this class be moved to?
+    public final class RSPos {
```
Tonodus: We have BlockVector for that.

src/main/java/net/glowstone/block/blocktype/BlockRedstoneTorch.java
```diff
@@ -0,0 +1,117 @@
...
+import net.glowstone.RSManager;
+import net.glowstone.block.GlowBlock;
+import org.bukkit.Material;
+import org.bukkit.block.BlockFace;
+
+import java.util.*;
+
+public class BlockRedstoneTorch extends BlockTorch {
+
+    public BlockRedstoneTorch(Material matType) {
```
Tonodus: You'd better use a boolean for on/off instead of every material.

src/main/java/net/glowstone/block/blocktype/BlockRedstoneTorch.java
```diff
@@ -0,0 +1,117 @@
...
+        if (face != BlockFace.UP && !isDirect) {
+            return false;
+        }
+
+        return true;
+    }
+
+    @Override
+    public boolean isRedSource(GlowBlock block) {
+        return true;
```
Tonodus, line 29: Always true, even if the torch is off?

kaenganxt, line 29: This is just used in RSManager to determine if this block should be added to the list of redstone sources to check regulary. Then the traceBlockPowerInit Method checks if this source is powered. And yes, the redstone torch always must be calculated, the reason for that is noted in the traceBlockPowerInit method.

gdude2002:
I know you're aware of it (and that this is WIP), but this has merge conflicts. You can check out what they actually are on my build server: http://bamboo.gserv.me/browse/GLOW-SRV-PR381-47

gdude2002:
http://bamboo.gserv.me/browse/GLOW-SRV-PR381-79

So close, yet.. not really that close. \o/

turt2live:
@kaenganxt Your WIP PR is looking pretty darn large. After many arguments, discussion, and overall complaining, we've decided that large PRs are nasty. My suggestion is to split the PR out into small chunks (yes, I know I sound like a hypocrite now) so we can review the bits better.

Logically the PR could be split into the following
- Basic redstone wire and torches
- Redstone blocks
- Buttons
- Levers
- Repeaters
- you get the idea...

turt2live:
~~@kaenganxt This WIP PR is looking pretty inactive. Please respond to my previous comment. ~~

Edit: Found the split PR

turt2live:
Closing this to clean up the PR list. I've found the split PRs thanks to the description.

PR #383, opened 2014-10-05
Fix downloads page not working due to jquery loading after the script

choyiny:
Make jquery load before the script by moving include footer on top.


choyiny:
Ah crap... new to pull requests, might have messed something up. It's just a single line change.
(commit 93d8deb)

bendem:
That would close all html tags before the last script tag... That's not valid html.

choyiny:
Hm... never thought of that. Lemme fix it now

bendem:
I'd just revert 1558d9b9da9060943c447541fe6f23ea3bd773a8 or remove the footer files.
There is no point in having the footer in another file if there is one of these file per actual html page.
(or maybe check in the footer if the file it is included in is the download page, but that'd overcomplicate stuff)

choyiny:
How to be able to make it automatically merge?

SpaceManiac:
Don't worry about it, it's probably just because it's against the gh-pages branch.

choyiny:
Done. Should be able to fix issue.

choyiny:
Use pull request #387 instead. closing.

Issue #384, opened 2014-10-05
Blocks with hardness set to 0 never really break, and doesn't drop.


jimmikaelkael:
All blocks with a value of 0 as Hardness never breaks on the server, it breaks on client side but not on server.
As a result all drops never get dropped.

I can take a few examples:
- Tall grass never drops seeds
- Pumpkin and melon stems never drop seeds.
- Torches never drop.

turt2live:
According to #629, `DiggingHandler.handle` doesn't call `DiggingMessage.FINISH_DIGGING` for these blocks.

Issue #379, opened 2014-10-02
Improved BlockType system


SpaceManiac:
Improve the BlockType system for 1.8 where variants based on properties are preferred and the type/data system is maintained solely in a backwards-compatibility capacity.

Pending initial information on SpongeAPI plans, for easier forward compatibility.

SpaceManiac:
See also #536.

Issue #372, opened 2014-09-29
Bedrock is breakable


mastercoms:
I'm making BlockBedrock in net.glowstone.block.blocktype, and I am making a new boolean method in BlockType called canBreak that BlockBedrock overrides. 
However, I'm not sure how to check canBreak from DiggingHandler in net.glowstone.net.handler.play.player.
I can only reference GlowBlock or Material in DiggingHandler, not BlockType.
I could add a boolean method in GlowBlock called canBreak that checks the material for bedrock but that just seems messy and inelegant to me.

Could someone familiarize me with the code if something like what I'm looking for exists, or give me some pointers on where to implement all of this if it needs to be coded.

SpaceManiac:
To get a hold of the BlockType for a Material, you can use `ItemTable.getBlock(Material)`.

Keep in mind that bedrock can be broken in creative mode, and only needs server-side dig prevention as an anti-cheat measure; a standard client will never 'finish' breaking a bedrock block. In favor of patching this special case, it may be worth solving dig timers in general.

mastercoms:
Thank you for that method.

canBreak will have a parameter, Player, so I will check the game mode of the player along with anything else if we need another check for other blocks. I guess I will expand this project to include anti-fast break measures too. Thanks for the suggestion.

SpaceManiac:
I'm not entirely certain a `canBreak` on BlockType is needed. It just seems a special case of dig time where a block has infinite dig time, something which doesn't affect creative by its nature.

mastercoms:
The reason why I suggested this was because of [Enderstone's implementation](https://github.com/BigTeddy98/Enderstone/blob/master/src/org/enderstone/server/blocks/BlockDefinitionBedrock.java#L38), but I guess the way Glowstone implemented blocks and digging doesn't fit with that.

PR #385, opened 2014-10-05
[WIP] Fix drops for blocks that should break instantly

jimmikaelkael:
Fixes issue #384.


jimmikaelkael:
While this fixes the problem, I'm not sure it is the best solution...

Don't we need to have Material hardnesses in the block registry instead ?
This could help to prevent a malicious client from digging too fast for example.

SpaceManiac:
I think I agree it would be beneficial to go ahead and work in material hardness into the registry somehow, which solves this and other weirdness with block hardness (e.g. #372) in one fell swoop, as well as finally adding that cheat prevention.

jimmikaelkael:
OK, I'll try to modify this PR to conform that.
Is a chained method call suitable to set the hardnesses ?

jimmikaelkael:
I've rebased to something like we discussed above.
Tell me if this is what we need, and I'll set other blocks hardness values.

SpaceManiac:
My main concerns are:
- `public setHardness` on BlockType - should somehow be finagled to not be public, IMO.
- `reg` in ItemTable returning the type. I think instead if `setHardness` stays like it is it should return `this` and the reg call should look like `reg(new XYZ().setHardness(2));`.
- `getHardness()` in GlowBlock. Maybe instead add a general-purpose `getBlockType` that looks up the `BlockType`, and call `.getBlockType().getHardness().`? This could then be used other places complicated `ItemTable` calls are currently used.

I need to give `ItemTable` a solid once-over for 1.8 anyways (see #379), so some of the issues above could be deferred until later.

Tonodus:
I think we should swap the hardness value along with other block-related values (f.e. blast resistance, light-level, and so on).
If we'd hardcode all this values in ItemTable that'd be hard to maintain.

jimmikaelkael:
This is not something that will change regularly... Most if not all will never change.

jimmikaelkael:
Rebased again with changes like we discussed above.
EDIT:
I'll add the other blocks hardness values and appropriate javadoc if this fix is convenient.

About setHardness being public in BlockType, it could be protected but then no way to call it from ItemTable directly...

jimmikaelkael:
Commited all blocks hardness values, for those already implemented in Glowstone of course.

Tonodus:
Setting the hardness value from ItemTable is not such a good approach IMO:

Blocks have many other values along with hardness, as said light-level, blast resistance, map-color, whether they are ignitable, and probably more I haven't in mind; the amount will increase with MC updates I assume.

The ItemTable isn't the right place to set all this values. They are all final, so it'd be good if there were no setter.
Instead I'd prefer a general class managing all this values and the mapping to the different Materials/Blocks.
Another approach would be public abstract getters for this values in BlockType; the value setting would be implemented then in the respective implementations of BlockType.

jimmikaelkael:
@Tonodus So something like what is done with recipes would be good to you ?

Tonodus:
@jimmikaelkael IMO the recipe solution is quite nice and could be used for this, too. So yes.

jimmikaelkael:
@Tonodus Anyway the 2nd approach you mentionned is not feasible IMO.
How would you handle the BlockType implementations when it's instanciated for several material then ?
I prefer the 1st approach.

turt2live:
It looks like this PR still has some discussion about approach going on. If there's a lot to do, please mark it as a WIP by prepending `[WIP]` to the title. If there is just some minor things to do, feel free to ping me using @turt2live when this PR can be thrown through testing. Thanks!

jimmikaelkael:
I think this should now be based on the new blocks branch.
EDIT: not sure... The content of commit b431f69 on master should be sufficient to start with.

SpaceManiac:
I wouldn't recommend it right now, the blocks branch isn't really final at all yet.

gdude2002:
Failing: http://bamboo.gserv.me/browse/GLOW-SRV-PR385-91

dequis:
Hm, the merge conflict is because 3166294b added `reg(Material.ENDER_PORTAL_FRAME, new BlockEnderPortalFrame());` to ItemTable.java, but since this PR changes pretty much all the block registry lines, sounds like you have to deal with this with every single change to that file while the PR is open.

Have fun with that!

(Actually I guess it's not a huge deal once you're used to it - it's a single conflict that looks ugly until you know what to do)

jimmikaelkael:
I'm rather busy with bonemeal/farming related PRs, but I'll solve this ASAP.

turt2live:
@jimmikaelkael psst. This doesn't merge thanks to the commit spam on master.

jimmikaelkael:
This PR is in standby as I've a lot of work with the Structures PR. At everyone: feel free to code a better alternative.

jimmikaelkael:
Closing this in favor of #541.

PR #387, opened 2014-10-06
Fix downloads page

HenrySlawniak:
This fix uses a tiny bit of front matter, and pushes the downloads page script into its own file.


HenrySlawniak:
This also keeps the foot.html consistent across all pages.

choyiny:
link with #383

PR #390, opened 2014-10-07
Add player visibility API 

### LINKED ISSUE #121: Visibility API
The `hidePlayer`, `showPlayer`, and `canSee(Player)` methods on `GlowPlayer` should be implemented. `canSee` checks should be performed when spawning players, displaying chat, etc.

The `canSee(GlowChunk.Key)` and `canSee(GlowEntity)` methods might want to be clarified as well.

Johni0702:
Resolves #121 
This implements the `hidePlayer`, `showPlayer` and `canSee(Player)` methods of `GlowPlayer` as well as performing necessary checks when sending the user list entries (tablist) and spawning the player entities.
Checks when sending chat (as suggested in the ticket) are not performed because (even though undocumented) no other Bukkit implementation does so.
Players will reappear after they reconnect (unless a plugin hides them again) which matches the behavior in CraftBukkit.

No changes were made to the `canSee(GlowChunk.Key)` and `canSee(GlowEntity)` methods as those are (even though similar named) not responsible for determining whether a chunk/entity is hidden from the player but rather whether they have been sent to the client.

This implementation hides players based on their UUIDs and therefore would also allow for hiding all other entities in the  the future.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -75,6 +75,11 @@ public final class GlowPlayer extends GlowHumanEntity implements Player {
     private final Set<GlowEntity> knownEntities = new HashSet<>();
 
     /**
+     * The entities that are hidden from the client.
+     */
+    private final Set<UUID> hiddenEntities = new HashSet<>();
+
+    /**
      * The chunks that the client knows about.
      */
     private final Set<GlowChunk.Key> knownChunks = new HashSet<>();
@@ -414,7 +419,7 @@ public final class GlowPlayer extends GlowHumanEntity implements Player {
                 continue;
             boolean withinDistance = !entity.isDead() && isWithinDistance(entity);
 
-            if (withinDistance && !knownEntities.contains(entity)) {
+            if (withinDistance && !knownEntities.contains(entity) && !hiddenEntities.contains(entity.getUniqueId())) {
                 knownEntities.add(entity);
                 for (Message msg : entity.createSpawnMessage()) {
                     session.send(msg);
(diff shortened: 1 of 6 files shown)


gdude2002:
Good stuff, oh-seven-oh-two. http://bamboo.gserv.me/browse/GLOW-SRV-PR390-52

ghost:
You should hide player from command auto-complete

Update:
@Johni0702 oh, sorry...

Johni0702:
This is already take care of by the auto-complete code which checks using `canSee(Player)` before suggesting a name.

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -1605,17 +1612,41 @@ public void sendWeather() {
...
+        }
+        session.send(UserListItemMessage.removeOne(player.getUniqueId()));
     }
 
     @Override
     public void showPlayer(Player player) {
+        Validate.notNull(player, "player cannot be null");
+        if (equals(player)) return;
+        if (!player.isOnline() || !session.isActive()) return;
+        if (!hiddenEntities.contains(player.getUniqueId())) return;
```
dequis, line 1630: Maybe turn these three into a single if?

dequis:
Tested with vanishnopacket, works as expected.

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -1605,17 +1612,41 @@ public void sendWeather() {
...
 
+        hiddenEntities.remove(player.getUniqueId());
+        session.send(UserListItemMessage.addOne(((GlowPlayer) player).getProfile()));
```
SpaceManiac: Should maybe use `GlowPlayer#getUserListEntry()` because it more accurately reflects the player's information.

SpaceManiac:
I might suggest renaming the existing `canSee(GlowChunk.Key)` and `canSee(GlowEntity)` to `canSeeChunk` and `canSeeEntity` respectively to avoid the need to cast things to `Player` in order to pass them to the correct `canSee` method.

Johni0702:
Added all three suggestions.
I didn't combine all three if clauses but only the two of them which are logically connected. The third one in `hidePlayer` is different from the one in `showPlayer` so I left those as they were.

SpaceManiac:
Merged in 51e0b340fb2caa36e9a2a552ab37fd3918986304.

Issue #392, opened 2014-10-08
Converter converts int to strings


greatman:
Basicly when glowstone converts the server.properties, the port parameter is converted to a string which makes that glowstone will never follow the glowstone.yml port.

PR #393, opened 2014-10-08
Convert strings to int and boolean when migrating from the server.properties

Johni0702:
Ticket: #392 
Convert strings read from the `server.properties` to `int`s or `boolean`s depending on the type of their default value.
If the value in the `server.properties` cannot be converted to an integer but is supposed to be one, prints out a warning and continues with the next property.


SpaceManiac:
Merged in 9a1a93b1eeabc8435c02f497f46feb1f3a46090a.

PR #402, opened 2014-10-10
Ender portals

Johni0702:
Ticket #325 

This adds the ability to place eye of ender into the ender portal frames. Once placed the surrounding blocks are scanned for a complete ender portal and if one is found the ender portal blocks are placed.
This also adds directional placement of ender portal frames and teleports the player/entities into the End / the Overworld on contact with an ender portal block.

Note that for a full portal to be formed, the last ender portal frame must point inwards in vanilla minecraft due to the way it detects the rest of the portal. The orientation of all other blocks is not checked. This is not the case with this PR, every orientation of the last portal frame will be sufficient to generate the portal.

The following events are now thrown for End portals:
- EntityPortalEvent
- EntityPortalEnterEvent
- EntityPortalExitEvent
- PlayerPortalEvent
- EntityCreatePortalEvent

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowEntity.java
@@ -1,6 +1,7 @@
 package net.glowstone.entity;
 
 import com.flowpowered.networking.Message;
+import net.glowstone.EventFactory;
 import net.glowstone.GlowChunk;
 import net.glowstone.GlowServer;
 import net.glowstone.GlowWorld;
@@ -11,9 +12,16 @@ import net.glowstone.util.Position;
 import org.apache.commons.lang.Validate;
 import org.bukkit.EntityEffect;
 import org.bukkit.Location;
+import org.bukkit.Material;
+import org.bukkit.World;
+import org.bukkit.World.Environment;
+import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
 import org.bukkit.entity.Entity;
 import org.bukkit.event.entity.EntityDamageEvent;
+import org.bukkit.event.entity.EntityPortalEnterEvent;
+import org.bukkit.event.entity.EntityPortalEvent;
+import org.bukkit.event.entity.EntityPortalExitEvent;
 import org.bukkit.event.player.PlayerTeleportEvent.TeleportCause;
 import org.bukkit.metadata.MetadataStore;
 import org.bukkit.metadata.MetadataStoreBase;
@@ -303,6 +311,28 @@ public abstract class GlowEntity implements Entity {
         if (ticksLived % (30 * 20) == 0) {
             teleported = true;
         }
+
+        if (hasMoved()) {
+            Block currentBlock = location.getBlock();
+            if (currentBlock.getType() == Material.ENDER_PORTAL) {
+                EventFactory.callEvent(new EntityPortalEnterEvent(this, currentBlock.getLocation()));
+                if (server.getAllowEnd()) {
+                    Location previousLocation = location.clone();
+                    boolean success;
+                    if (getWorld().getEnvironment() == Environment.THE_END) {
+                        success = teleportToSpawn();
+                    } else {
+                        success = teleportToEnd();
+                    }
+                    if (success) {
+                        EntityPortalExitEvent e = EventFactory.callEvent(new EntityPortalExitEvent(this, previousLocation, location.clone(), velocity.clone(), new Vector()));
+                        if (!e.getAfter().equals(velocity)) {
+                            setVelocity(e.getAfter());
+                        }
+                    }
+                }
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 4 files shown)


rbrick:
:+1:

turt2live:
This PR does add additional functionality that CraftBukkit did not have, such as firing the `EntityCreatePortalEvent`.

turt2live:
Noted issue (out of scope of this PR): Breaking the ender portal shows the wrong block particles

Johni0702:
Wrong block break particles of the ender portal block seems to be a vanilla minecraft issue.

src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -427,6 +457,54 @@ public boolean hasRotated() {
         return Position.hasRotated(location, previousLocation);
     }
 
+    /**
+     * Teleport this entity to the spawn point of the main world.
+     * This is used to teleport out of the End.
+     * @return {@code true} if the teleport was successful.
+     */
+    public boolean teleportToSpawn() {
```
SpaceManiac: Do this and the other method need to be public? Would `protected` suffice?

src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -303,6 +311,28 @@ public void pulse() {
         if (ticksLived % (30 * 20) == 0) {
             teleported = true;
         }
+
+        if (hasMoved()) {
+            Block currentBlock = location.getBlock();
+            if (currentBlock.getType() == Material.ENDER_PORTAL) {
+                EventFactory.callEvent(new EntityPortalEnterEvent(this, currentBlock.getLocation()));
```
SpaceManiac, line 318: Is it intended behavior to call this event even if the end is disabled?

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -1160,6 +1161,55 @@ public boolean teleport(Location location, TeleportCause cause) {
...
+        PlayerPortalEvent event = EventFactory.callEvent(new PlayerPortalEvent(this, location.clone(), target, null));
+        if (event.isCancelled()) {
+            return false;
+        }
+        target = event.getTo();
+
+        spawnAt(target);
+        teleported = true;
+
+        awardAchievement(Achievement.THE_END);
```
SpaceManiac: These achievements would be awarded even if their prerequisites had not been; perhaps a helper method that checks the achievement's parent before awarding it would be helpful here.

dequis: Sounds like an issue of awardAchievement, out of scope for this PR

turt2live: From the JavaDocs:

`Awards the given achievement and any parent achievements that the player does not have.`

SpaceManiac: I'll take care of it, then.

src/main/java/net/glowstone/entity/GlowEntity.java
turt2live, line 318: CraftBukkit behaviour is no as far as my testing goes.

Johni0702, line 318: @turt2live The event does get called repeatedly using Spigot (#1647) and CraftBukkit (1.7.2-R0.2), I just tried that again. I don't have a recent CraftBukkit build, so could you please verify your tests and/or notify me on IRC (if I'm there).
Testing procedure (using [ScriptCraft](http://scriptcraftjs.org/)):
1. Disable `settings.allow-end` in `bukkit.yml`
2. Start Server with ScriptCraft enabled
3. Type `/js events.on(org.bukkit.event.entity.EntityPortalEnterEvent, function(e){echo("a");});`
4. Stand in the end portal
5. Chat gets spammed with `a` (otherwise event wouldn't get called).

src/main/java/net/glowstone/entity/GlowPlayer.java
Johni0702: Vanilla only grants the achievement if the player has the parent achievement ("Into Fire" and "The End?" in this case).
So there should probably be a second method for awarding an achievement only if the players already has the parent achievement.
That method should then be used internally on every occasion where we want to update the progress of a player's achievements.
I just cannot think of a good name for that method. My only idea `awardAchievementIfHasParent` sounds somewhat bulky.

src/main/java/net/glowstone/entity/GlowEntity.java
turt2live, line 318: Hmm. Something went wrong with my testing (I seem to be good at that lately). It indeed does spam the hell out of chat. 

Question should be rephrased as `Should Glowstone implement the same bug?`

src/main/java/net/glowstone/entity/GlowPlayer.java
turt2live: I would write an overloaded method:

`void awareAchievement(Achievement ach, boolean awardParents);`

src/main/java/net/glowstone/entity/GlowEntity.java
Johni0702, line 318: I don't think this should be considered a bug. From JavaDocs:
`Called when an entity comes into contact with a portal`
The event is for touching the portal, not for using it, so it should still fire even if the End is disabled. It might also be used by some teleport plugins to create custom portals which are supposed to work as well with End disabled.

turt2live, line 318: I suppose so, yes. Just another instance of a vague event in Bukkit >.>

Johni0702:
Reduced access levels as suggested.
Added a method `awardAchievement(Achievement achievement, boolean awardParent)` which if called with `true` acts as the javadoc of `awardAchievement(Achievement achievement)` describes. And if called with `false` only awards the achievement if the player already has the parent achievement.

bendem:
Wouldn't it be useful if that method returned true or false since it can fail or not?

turt2live:
Changes look safe enough to be re-marked for pulling

SpaceManiac:
Concerns:
- It looks like end portals are created even if there's something obstructing their insides. Is this intended?
- The `GlowPlayer` overrides of the teleport methods use `spawnAt` instead of `teleport`. This means `PlayerTeleportEvent` is not called. Is this intended?

Johni0702:
> It looks like end portals are created even if there's something obstructing their insides. Is this intended?

Yes, it is. This matches vanilla minecraft behavior.

> The GlowPlayer overrides of the teleport methods use spawnAt instead of teleport. This means PlayerTeleportEvent is not called. Is this intended?

Yes, it is. Instead the [PlayerPortalEvent](http://jd.bukkit.org/dev/apidocs/org/bukkit/event/player/PlayerPortalEvent.html) is called which extends the `PlayerTeleportEvent`.

SpaceManiac:
Merged in 3166294bf0197752ff3cc1f5598d1426e9959e9d.

PR #413, opened 2014-10-12
Implemented basic damage methods

PaulBGD:
After a bit of testing, it appears to work fine. With players it does properly update them, and give them the respawn screen when they die.

https://github.com/GlowstoneMC/Glowkit/pull/33 is required

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowLivingEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowLivingEntity.java
@@ -1,20 +1,40 @@
 package net.glowstone.entity;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
 import net.glowstone.constants.GlowPotionEffect;
+import org.bukkit.Bukkit;
+import org.bukkit.EntityEffect;
 import org.bukkit.Location;
 import org.bukkit.Material;
+import org.bukkit.Sound;
 import org.bukkit.block.Block;
-import org.bukkit.entity.*;
+import org.bukkit.entity.Arrow;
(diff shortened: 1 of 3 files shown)


turt2live:
EDITED: I'M GOING TO FIX THIS THING EVENTUALLY

ZephireNZ:
@turt2live Your PR bot derped... Again.

turt2live:
@ZephireNZ I know >.> Fixed already.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -357,6 +442,13 @@ public void setLastDamage(double damage) {
     }
 
     ////////////////////////////////////////////////////////////////////////////
+    // Abstract inventory methods
+
+    public void dropItems() {
```
turt2live: I know I said otherwise in IRC, but instead of making an empty method for this, just leave a `// todo` where it would be called. Also add a `// todo` somewhere around there for "death event".

gdude2002:
Good stuff. http://bamboo.gserv.me/browse/GLOW-SRV-PR413-83

gdude2002:
Tch. What happened? http://bamboo.gserv.me/browse/GLOW-SRV-PR413-87

gdude2002:
Passed. I think. http://bamboo.gserv.me/browse/GLOW-SRV-PR413-88

turt2live:
Looks good enough. Death is buggy as hell, but that's for another PR. 

Drowning works :)

SpaceManiac:
Is the Glowkit PR really needed, or can the new methods just be available Glowstone-side and everything still work? I'm not really sure if these methods need to be available to plugins.

PaulBGD:
I personally believe they do need to be there, so that plugins can damage entities accordingly.

SpaceManiac:
Merged in 149021fd92a5bac413d251efc0bcf4adf53ec053.

Issue #377, opened 2014-10-01
Query support


SpaceManiac:
http://wiki.vg/Query
(low-priority)

PR #414, opened 2014-10-12
Query support

Johni0702:
Ticket: #377 
This adds query support as described on http://wiki.vg/Query
Tools for testing can be found on the bottom of said page.
Unit tests are available on the [`query-test`](https://github.com/Johni0702/Glowstone/commits/query-test) branch. I didn't want to include them in the PR yet because they depend on mockito and powermock.

### DIFF
--- a/src/main/java/net/glowstone/net/query/QueryServer.java
+++ b/src/main/java/net/glowstone/net/query/QueryServer.java
@@ -1,0 +1,127 @@
+package net.glowstone.net.query;
+
+import io.netty.bootstrap.Bootstrap;
+import io.netty.channel.ChannelFuture;
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.nio.NioEventLoopGroup;
+import io.netty.channel.socket.nio.NioDatagramChannel;
+import net.glowstone.GlowServer;
+import org.bukkit.scheduler.BukkitRunnable;
+
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Random;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Implementation of a server for the minecraft server query protocol.
+ * @see <a href="http://wiki.vg/Query">Protocol Specifications</a>
+ */
+public class QueryServer {
+    /**
+     * The {@link EventLoopGroup} used by the query server.
+     */
+    private EventLoopGroup group = new NioEventLoopGroup();
+
+    /**
+     * The {@link Bootstrap} used by netty to instantiate the query server.
+     */
+    private Bootstrap bootstrap = new Bootstrap();
+
+    /**
+     * Instance of the GlowServer.
+     */
+    private GlowServer server;
+
+    /**
+     * Maps each {@link InetSocketAddress} of a client to its challenge token.
+     */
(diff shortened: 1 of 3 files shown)


gdude2002:
_ding sound_

http://bamboo.gserv.me/browse/GLOW-SRV-PR414-83

SpaceManiac:
Merged in e854d50.

src/main/java/net/glowstone/net/query/QueryServer.java
```diff
@@ -0,0 +1,127 @@
...
+public class QueryServer {
+    /**
+     * The {@link EventLoopGroup} used by the query server.
+     */
+    private EventLoopGroup group = new NioEventLoopGroup();
+
+    /**
+     * The {@link Bootstrap} used by netty to instantiate the query server.
+     */
+    private Bootstrap bootstrap = new Bootstrap();
```
mastercoms, line 31: Any reason why just Bootstrap was used instead of ServerBootstrap?

Johni0702: Because the query server speaks UDP only and for UDP there is only ever one channel regardless of the number of clients.
The Netty docs mentioning it: http://netty.io/4.1/api/io/netty/bootstrap/Bootstrap.html (second paragraph)
See also: http://stackoverflow.com/a/37681245

mastercoms: Ok thank you!

mastercoms: I looked more into it and I don't see why ServerBootstrap wouldn't allow for that.

mastercoms: Would it still be possible to use ServerBootstrap?

Johni0702: I don't think so.
ServerBootstrap is a ["Bootstrap sub-class which allows easy bootstrap of ServerChannel"](http://netty.io/4.0/api/io/netty/bootstrap/ServerBootstrap.html). However none of the [DatagramChannel](http://netty.io/4.0/api/io/netty/channel/socket/DatagramChannel.html) implementations (and therefore specifically [NioDatagramChannel](http://netty.io/4.0/api/io/netty/channel/socket/nio/NioDatagramChannel.html)) implement the ServerChannel interface.

Is there a problem with using Bootstrap?

mastercoms: No, I was just trying to clean up the server code, specifically the bind code. I wanted to make a hierarchy but I guess I will have to add another layer for socket and datagram.

mastercoms: I am also having trouble with your test code in the query-test branch. I've uploaded the code I have [here](https://github.com/GlowstoneMC/Glowstone/blob/netserver/src/test/java/net/glowstone/net/QueryTest.java) and I would be very grateful if you could give me any pointers on how to get the test to pass.

Johni0702: With [this](https://i.johni0702.de/0001-Fix-server-query-test.patch) patch applied, all tests pass.

mastercoms: thanks so much!

Issue #411, opened 2014-10-12
Locking up the main thread causes disconnection inconsistencies


turt2live:
When the main thread locks up (change a few million blocks with WorldEdit), players are able to still talk to each other but the rest of the server is frozen. That's logical and makes sense. What doesn't is that when the main thread resumes (WE task done), all players are booted due to a timeout error. They are actually kicked for a timeout error and not just a socket timeout (from the looks of it). 

As a player, I'd expect one of the following to occur:

1) Be disconnected part-way through the task
2) Remain connected after the task (ideal)
3) Not have the entire server freeze due to a task (async networking, less ideal for code/performance concerns)

ghost:
Or other thread for we...

bendem:
@ensirius Not sure Glowstone is ready for async world modifications (pretty sure it isn't)...

turt2live:
Closing this to clean up my issues list on Github. If this is still an issue, please port it over to https://github.com/GlowstoneMC/Glowstone

PR #415, opened 2014-10-12
Prevent time out when main thread is blocked

Johni0702:
Ticket: #411 
This prevents the player from getting kicked during or after an expensive operation in the main thread by moving the necessary checks into another thread which will never be blocked by any expensive operations.
This thread would also be the place where one would put a potential future watchdog system for the main thread.
Testing can be done with any expensive main-thread-locking plugin such as WorldEdit or ScriptCraft (`/js java.lang.Thread.sleep(30000)`) by blocking the main thread for more than 30 seconds.

---

_Edited by turt2live_

**Related Links:**
- Ticket: #411

### DIFF
--- a/src/main/java/net/glowstone/net/GlowSession.java
+++ b/src/main/java/net/glowstone/net/GlowSession.java
@@ -378,6 +378,12 @@ public final class GlowSession extends BasicSession {
             GlowServer.logger.info("[" + address + "] kicked: " + reason);
         }
 
+        if (timedOut) {
+            quitReason = "timed out";
+            // consider this session gone
+            onDisconnect();
+        }
+
         if (quitReason == null) {
             quitReason = "kicked";
         }
@@ -413,8 +419,8 @@ public final class GlowSession extends BasicSession {
         }
 
         // connection timed out, we should disconnect them
-        if (timedOut) {
-            disconnect("Timed out");
+        if (timedOut && player != null) {
+            disconnect("Timed out", true);
         }
     }
 
--- a/src/main/java/net/glowstone/net/GlowSession.java
+++ b/src/main/java/net/glowstone/net/GlowSession.java
@@ -107,6 +107,12 @@ public final class GlowSession extends BasicSession {
     private int writeTimeoutCounter = 0;
 
     /**
+     * Whether to kick the player the next time its updated. This is set to {@code true}
+     * when a timeout occurs in order for the player to be kicked from the main thread.
+     */
+    private boolean timedOut = false;
+
+    /**
      * The player associated with this session (if there is one).
      */
     private GlowPlayer player;
@@ -389,9 +395,6 @@ public final class GlowSession extends BasicSession {
      * Pulse this session, performing any updates needed.
      */
     void pulse() {
-        readTimeoutCounter++;
-        writeTimeoutCounter++;
-
         // drop the previous placement if needed
         if (previousPlacementTicks > 0 && --previousPlacementTicks == 0) {
             previousPlacement = null;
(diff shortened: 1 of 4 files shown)


gdude2002:
Looks good. http://bamboo.gserv.me/browse/GLOW-SRV-PR415-83

src/main/java/net/glowstone/net/GlowSession.java
```diff
@@ -107,6 +107,12 @@
     private int writeTimeoutCounter = 0;
 
     /**
+     * Whether to kick the player the next time its updated. This is set to {@code true}
```
mastercoms: Grammar Nazi alert: it's, not its

turt2live:
@Johni0702 this currently has a major bug where the console can (and will) be spammed with "Timed out" kick messages if the player is sent packets after they have been kicked once.

Scenario: I started 150 bots on a server and did a 17.3 million block change in WorldEdit. While WorldEdit was processing the change, I disconnected 40 bots using a hard disconnect (therefore leaving the server connection open). Once WorldEdit stopped saturating the main thread, the block updates were distributed and all 40 of those clients got a few million messages in the console each. After the kick message spam stopped, I did a smaller change (5000 blocks) and once again got the same spam message.

Here's an excerpt from the log on a smaller test to verify that this PR was at fault:

```
14:24:34 [INFO] testBot_113 kicked: Timed out
14:24:34 [INFO] testBot_44 kicked: Timed out
14:24:34 [INFO] testBot_116 kicked: Timed out
14:24:34 [INFO] testBot_113 kicked: Timed out
14:24:34 [INFO] testBot_44 kicked: Timed out
14:24:34 [INFO] testBot_116 kicked: Timed out
14:24:34 [INFO] testBot_113 kicked: Timed out
14:24:34 [INFO] testBot_44 kicked: Timed out
14:24:34 [INFO] testBot_116 kicked: Timed out
14:24:34 [INFO] testBot_113 kicked: Timed out
14:24:34 [INFO] testBot_44 kicked: Timed out
14:24:34 [INFO] testBot_116 kicked: Timed out
14:24:34 [INFO] testBot_113 kicked: Timed out
14:24:34 [INFO] testBot_44 kicked: Timed out
14:24:34 [INFO] testBot_116 kicked: Timed out
14:24:34 [INFO] testBot_113 kicked: Timed out
14:24:34 [INFO] testBot_44 kicked: Timed out
```

turt2live:
@Johni0702 New issue: Sending ~9 million block changes to 153 players caused the clients to timeout from the server (normal), but the server fell into a full saturation of all resources well after the block changes were sent (I was peeking at the server while doing this). Looks like there's an infinite loop somewhere.

Johni0702:
@turt2live I cannot seem to find any reason for the full saturation of all resources. Neither can I reproduce it as I do not have the required hardware to do so. Would you mind to test it again at some point and get me a thread dump of when it happens? When doing so, please also note down after how many seconds stuff happens (WE operation duration, first timeouts, seconds after WE operation till the thread dump, etc.)  as that might help figuring out this problem.

turt2live:
I'm putting this back into `Needs Testing` so I can attempt to get those thread dumps.

dequis:
Will probably be replaced by #189

turt2live:
I'm assigning this to @SpaceManiac to review simply because I have little knowledge about netty. Ideally a comparison between this and #189 should be done in the meantime ( @coelho ?)

coelho:
@turt2live My PR completely obsoletes the old system that this PR relies on.

Johni0702:
I'd personally also prefer @coelho's PR as it's far more nettyish and less complicated for doing the same job. If his PR doesn't have any mandatory issues feel free to close this one, it's not even worth looking into compared to coelho's.

turt2live:
Closing in favour of #189.

PR #374, opened 2014-10-01
[WIP] Add bonemeal interaction

jimmikaelkael:
I'm creating this PR with the hope we can discuss the commit comment below.

EDIT: I'm extending this PR since I've created IBlockGrowable interface.
### Work to be done:
- [X] add ItemDye (rightClickBlock() override)
- [X] add ItemHoe (rightClickBlock() override)
- [X] add ItemSeeds (rightClickBlock() override)
- [X] add ItemFoodSeeds (rightClickBlock() override)
- [X] add IBlockGrowable
- [x] add isFertilizable() and canGrowWithChance() methods to IBlockGrowable
- [X] add BlockPlant
- [X] add BlockStem (extend BlockPlant, replaces BlockMelonStem & BlockPumpkinStem)
- [X] add BlockCarrot (extend BlockCrops)
- [x] add BlockCocoa
- [X] add BlockCrops (extend BlockPlant)
- [X] add BlockDeadBush (extend BlockPlant)
- [X] modify BlockDoublePlant to extend BlockPlant
- [X] add BlockGrass
- [x] add BlockMushroom (do not extend BlockPlant)
- [X] add BlockPotato (extend BlockCrops)
- [x] add BlockPumpkin (control facing direction in placeBlock override)
- [X] modify BlockSapling to extend BlockPlant
- [X] modify BlockTallGrass to extend BlockPlant
- [X] add huge mushrooms generation
- [x] add cocoa beans interaction
- [x] modify the particles spawned on fertilized block (only if IBlockGrowable instance of, and IBlockGrowable.isFertilizable() returns true)
- [x] call IBlockGrowable.canGrowWithChance() method before to call IBlockGrowable.fertilize() 
- [x] remove the random chance to grow in BlockSapling.fertilize() method
- [x] put generateHugeMushroom() method in it's own generator, HugeMushroomGenerator (extends TreeGenerator)
- [x] move World.generateTree() content to TreeGenerator
- [x] try to implement a basic block  random tick system
- [x] trigger BlockGrowEvent where needed
- [x] trigger StructureGrowEvent where needed
- [x] trigger BlockSpreadEvent where needed
- [x] check styles and javadoc (concerning the code in this PR)
- [x] provide some test plugins: they will be needed to test BlockGrowEvent, StructureGrowEvent and BlockSpreadEvent
### Related issues / PR:
- #206
- #385 
- https://github.com/GlowstoneMC/Glowkit/pull/27
- https://github.com/GlowstoneMC/Glowkit/pull/29
[...]

PR #426, opened 2014-10-13
Add a script that fetches glowkit PRs for travis

dequis:
Also reorganize .travis.yml slightly (actual build stuff goes in the script step, preparations go in the install step)

The script uses the github API to fetch the pull request body (which is needed to find the relevant PR links).

Since the github api has very strict rate limits for unauthenticated users, I've included an [oauth personal access token](https://github.com/settings/tokens/new) with scope for public repositories only. This means that compared to unauthenticated access, it's only useful to bypass the rate limit

I'd still recommend using the travis encryption features for the token, just in case. I'd like to revoke mine as soon as testing is done.

Also note that using the encryption features means the script won't work in other repos, but this isn't an issue since it's also hardcoded to fetch PRs from Glowstone/Glowstone and Glowstone/Glowkit.

Did some testing for it [here](https://travis-ci.org/dequis/Glowstone/pull_requests) and [here](https://travis-ci.org/dequis/Glowstone/builds). The commit labeled "TESTING SHIT" (now reverted) calls the script three times with hardcoded ids instead of `$TRAVIS_PULL_REQUEST`. These PRs are:
- #374 - no glowkit PRs required
- #402 - two PRs ~~required~~ mentioned, ignored by the script. The author of that PR said they aren't needed.
- #413 - one PR required, detected correctly by the script and built successfully, but in the end the build fails because the glowstone side doesn't have the required changes. For the purposes of my script, this means it works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,22 @@
 language: java
 jdk: oraclejdk7
 
-# Compile and package JAR and set build properties
-install: ./gradlew clean build remap
+
+
+install:
+
+  # Required for fetch_glowkit_pr.sh
+  - "sudo apt-get install jq"
+
+  # Try to fetch required glowkit PRs, if any.
+  - "./etc/fetch_glowkit_pr.sh $TRAVIS_PULL_REQUEST || echo 'Glowkit build failed - trying to continue anyway'"
+
+  # Fetch gradle stuff, so it doesn't appear in the script step
+  - "./gradlew clean"
+
+script:
+  # Compile and package JAR and set build properties
+  - "./gradlew build remap"
 
 after_success:
   # Check if commit is not a pull request, if repo is official, and branch is master, generate and deploy artifacts and reports
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@
 language: java
 jdk: oraclejdk7
 
-
-
 install:
 
   # Required for fetch_glowkit_pr.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ jdk: oraclejdk7
 install:
 
   # Required for fetch_glowkit_pr.sh
-  - "sudo apt-get install jq"
+  - "[[ $TRAVIS_PULL_REQUEST == false ]] || sudo apt-get install jq"
 
   # Try to fetch required glowkit PRs, if any.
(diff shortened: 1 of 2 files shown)


turt2live:
There is one failing case: `GlowstoneMC/Glowkit#000`. That format is used and is not parsed.

dequis:
Added the requested changes, plus checking if the linked PRs are issues/closed PRs.

If you wish to test the parsing of PR body without submitting actual PRs, comment out the `pr_body=...` line and use something like this instead:

```
pr_body='GlowstoneMC/Glowkit#30 GlowstoneMC/Glowkit#25 GlowstoneMC/Glowkit#12 GlowstoneMC/Glowkit#26 GlowstoneMC/Glowkit#0'
```

turt2live:
Leaving this to @SpaceManiac to decide whether or not it should be included

dequis:
Added another commit that should make it easier to switch to encrypted tokens on the travis side - just comment out (or remove) the line that sets `GITHUB_ACCESS_TOKEN`, and provide it externally.

The script can continue to work with the access token removed when using it outside of travis, but the rate limit of 60 requests per hour makes it useless for anything else.

SpaceManiac:
A couple of questions:
- Would it work to move `sudo apt-get install jq` the script so it isn't installed on non-PR builds?
- (Somewhat unrelated, but while we're on the Travis script) I understand using the wrapper is the recommended way to do things, but downloading Gradle now takes up over half the build. Is it reasonable to use the system `gradle` instead?

dequis:
> Would it work to move sudo apt-get install jq the script so it isn't installed on non-PR builds?

I wanted to keep the script as travis-agnostic as possible, but i'll add some real PR checks to the install script.

> Is it reasonable to use the system gradle instead?

Yeah, I think so. Will test.

dequis:
System gradle works: https://travis-ci.org/dequis/Glowstone/builds/38678302 (the build ends as a failure because it's using outdated glowstone with a newer glowkit, but meh)

Also seems that this reduces the build time by 40 seconds \o/

dequis:
Dammit @turt2live.

_pokes @SpaceManiac_

dequis:
1 min 38 sec (this build)
vs
2 min 14 sec (previous successful build)

:D

SpaceManiac:
Merged in 25dc62aa229709018ff821bb9fb9c83c318ae667.

ZephireNZ:
Cannot merge. GG

dequis:
good job @turt2bot

PR #433, opened 2014-10-14
Basic farming

jimmikaelkael:
PR splitting as discussed in #374.

This PR allow to plant wheat, melon and pumpkins seeds, nether warts as well as cocoa beans.
This seems to be ready for review.

PR #432, opened 2014-10-14
Bonemeal interaction

jimmikaelkael:
This handles the bonemeal interaction with:
- Carrots and potatoes
- Cocoa
- Crops
- DoublePlants (the flowers only, as expected)
- Grass
- Tall grasses (normal and fern)
- Melon and Pumpkin stems
- Brown and red mushrooms
- Saplings (I worked around the bukkit bug with saplings MaterialData)
#### Progress
- [x] add tree generation
- [x] add huge mushroom generation
- [x] provide a test plugin for BlockGrowEvent and StructureGrowEvent
#### Test Plugin:
- jar: https://drive.google.com/file/d/0B0gnjB7PN1h3X0VKRnlOMmozZUU/view?usp=sharing
- source available here: https://gist.github.com/jimmikaelkael/b7aac6c8c27a8dd2aed5

---

_Edited by turt2live_

**Related Links:**
- Ticket: #206 
- Ticket: #325 
- Waiting for #492 (or similar)
- Waiting for #289 (or similar)

PR #430, opened 2014-10-14
Add huge mushrooms and simple tree generator

jimmikaelkael:
This is corresponding to PR splitting as requested in #374.

This adds a complete huge mushroom generator. The old tree generator glowstone used is moved to TreeGenerator but it still the same.

Tree generation from GlowWorld.generateTree() method calls the StructureGrowEvent as it is expected.
#### Todo
- [x] rewrite a better generic tree generator (generates oak tree, small jungle trees and cocoa trees)
- [x] add birch / tall birch tree generator
- [x] add spruce tree generator
#### Test Plugin
- source avalable here: https://gist.github.com/jimmikaelkael/3d3448223c64870bbecd
- jar: https://drive.google.com/file/d/0B0gnjB7PN1h3SlEwZDFMdzdlV1U/view?usp=sharing

PR #434, opened 2014-10-14
Farming blocks placement check and drops

jimmikaelkael:
PR splitting as expected in #374.
Seems it's ready to be reviewed.


src/main/java/net/glowstone/block/blocktype/BlockCarrot.java
```diff
@@ -0,0 +1,26 @@
...
+
+import net.glowstone.block.GlowBlock;
+
+import org.bukkit.CropState;
+import org.bukkit.Material;
+import org.bukkit.inventory.ItemStack;
+
+public class BlockCarrot extends BlockCrops {
+
+    private final Random random = new Random();
```
turt2live: Move all instances of `Random` to `BlockType` as a protected static field. Too many things access a random object for it to be justified to create several copies of the same seeded random.

jimmikaelkael: This is very true. That was originally in m code comments in the huge PR.
Will change.

jimmikaelkael:
@turt2live update with changes suggested above.

src/main/java/net/glowstone/block/blocktype/BlockPotato.java
```diff
@@ -0,0 +1,28 @@
...
+import org.bukkit.CropState;
+import org.bukkit.Material;
+import org.bukkit.inventory.ItemStack;
+
+public class BlockPotato extends BlockCrops {
+
+    @Override
+    public Collection<ItemStack> getDrops(GlowBlock block) {
+        if (block.getData() >= CropState.RIPE.ordinal()) {
+            if (random.nextInt(100) < 2) {
```
ZephireNZ, line 18: Interesting choice for random. Couldn't this simply be random.nextInt(50) == 0

Currently it's 2 in 100 - either 0 or 1 will return true - which can be simplified to 1 in 50

turt2live, line 18: `The poisonous potato is a rare drop when harvesting (destroying) potato plants: a fully grown plant has a 2% chance of dropping one in addition to the 1–4 regular potatoes.`

Simplifying this would make it less readable.

jimmikaelkael, line 18: @turt2live that's exactly what I used as a reference :)

src/main/java/net/glowstone/block/blocktype/BlockMushroom.java
```diff
@@ -0,0 +1,33 @@
...
+        this.mushroomType = mushroomType;
+    }
+
+    @Override
+    public boolean canPlaceAt(GlowBlock block, BlockFace against) {
+        final GlowBlock belowBlock = block.getRelative(BlockFace.DOWN);
+        final Material type = belowBlock.getType();
+        if (type.equals(Material.GRASS)
+                || (type.equals(Material.DIRT) && belowBlock.getData() != 2)) {
+            // uncomment later to stop mushroom growth where there's too much light
```
turt2live: Might as well uncomment now considering it's a very small change and already here

src/main/java/net/glowstone/block/blocktype/BlockDeadBush.java
```diff
@@ -0,0 +1,31 @@
...
+
+import net.glowstone.block.GlowBlock;
+
+import org.bukkit.Material;
+import org.bukkit.block.BlockFace;
+import org.bukkit.inventory.ItemStack;
+
+public class BlockDeadBush extends BlockPlant {
+
+    private final Collection<ItemStack> emptyStack = Collections.unmodifiableList(Arrays.asList(new ItemStack[0]));
```
turt2live, line 15: I almost feel as though this should be in the base class as well... Might make my own change for this however.

src/main/java/net/glowstone/block/blocktype/BlockMushroom.java
jimmikaelkael: True, if left commented mushrooms will possibly spread almost everywhere, could be long but still.

turt2live: Well, this PR doesn't have ticking so they wouldn't spread. But if ticking were to be added and this forgotten, then yea... mushroom worlds.

src/main/java/net/glowstone/block/blocktype/BlockDeadBush.java
jimmikaelkael, line 15: If you're really decided then I can do it now, as anyway I'll use BlockPlant in other farming related code.

turt2live, line 15: Not quite the change I meant, but okay. The moving is out of scope of the PR anyways.

jimmikaelkael, line 15: Ah sorry ^^
But seems what I understood makes sense.
EDIT:
I had a very bad idea: saplings won't drop anymore, I'll revert last commit.

src/main/java/net/glowstone/block/blocktype/BlockStem.java
```diff
@@ -0,0 +1,44 @@
...
+    private Material seedsType;
+
+    public BlockStem(Material plantType) {
+        if (plantType.equals(Material.MELON_STEM)) {
+            fruitType = Material.MELON_BLOCK;
+            seedsType = Material.MELON_SEEDS;
+        } else if (plantType.equals(Material.PUMPKIN_STEM)) {
+            fruitType = Material.PUMPKIN;
+            seedsType = Material.PUMPKIN_SEEDS;
+        }
```
turt2live: For sanity, `else throw new IllegalArgumentException("insert message");` with the proper formatting :3

src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -23,12 +23,14 @@
...
 import java.util.Collections;
 import java.util.List;
+import java.util.Random;
 
 /**
  * Base class for specific types of blocks.
  */
 public class BlockType extends ItemType {
 
+    protected static final Random random = new Random();
```
turt2live, line 33: Because someone else may ask why this was done, I'm re-commenting on it:

I asked for this to be done as the random object was being used in multiple subclasses. There is a larger fix to do with respect to this, but this solves the immediate problem with the PR where multiple states could be seeded with the same random.

Hopefully doing this will solve some future problems with not-so-random things.

turt2live:
~~This is looking good, however I cannot place and seeds. The item is consumed and the plant shows up for a moment, but then it disappears. Please fix this and we'll likely be on the track to pulling asap!~~

I'm an idiot. https://github.com/GlowstoneMC/Glowstone/pull/433

turt2live:
@jimmikaelkael A lot of these are already covered in your other PR, #433. Because #433 would likely be pulled first, I would recommend making this PR an extension of that PR (therefore on hold for that PR) where other plants can be implemented here.

jimmikaelkael:
@turt2live huh ? You meant #432 no ?
I'd rather say it's an extension of #432 unless I'm missing something.
But yes place this on hold.

jimmikaelkael:
Honestly I think this and #432 should anyway be placed on hold until #430 is pulled.
This PR could also have improvements if #442 is pulled first: 
- check if jungle log is destroyed to drop cocoa if it's attached
- check if soil is destroyed to drop crops

turt2live:
@jimmikaelkael I'm just confused now. When I get the chance I'm going to find all of your PRs and prioritize them. I think I might have asked for too much splitting XD (don't recombine them, I'll just clean up the mess I've asked for)

jimmikaelkael:
@turt2live Should I merge this PR in #432 ?

turt2live:
@jimmikaelkael I still have to figure out the order of your various PRs. When I do that later today, I'll let you know.

jimmikaelkael:
Closing this as the content related to bonemeal interaction has been merged in #432.
I'll open antoher PR later for nether warts and dead bushes.

PR #249, opened 2014-09-10
Block physics

Tonodus:
Some basic block physics. Please give feedback to the structure.

Edit for clarification:
This PR provides the possibility to add block physics to certain BlockTypes, so that developers developing new BlockTypes can easily override the methods and program the related physic stuff.
It doesn't add all physics itself, only falling blocks (gravel & sand) and the going out of fire if the block below is broken.

PR #437, opened 2014-10-14
Implemented door placement and door handling (See #119)

BarelyAPrincess:
An almost completed implementation of door placement and handling.  I used the temporary method onDestroy which will need to be mapped to a method in the BlockType class, so the door is properly removed when destroyed.  I also added redstone support which can't be tested and implemented until the redstone framework is added.

This is my first time contributing to a community project, I hope I did it right.


ZephireNZ:
IMO I'd say remove the redstone stuff until redstone is merged. Otherwise this might get overlooked when redstone is added, or - more likely - redstone will change and this will be broken.

turt2live:
Honestly this should be split up anyways: Doors and redstone interaction with doors

BarelyAPrincess:
What's your thoughts on adding the onDestroy method to the BlockType class? I was able to make it all work perfectly by calling onDestroy method from the end of net.glowstone.net.handler.play.player.DiggingHandler class.

I can write up a Pull Request for that too if you think that should work.

turt2live:
@ChioriGreene This PR is still marked as a WIP. If this is ready for review, please remove `[WIP]` from the title. Otherwise please respond stating that this is still in progress.

Additionally, this PR cannot be merged - please resolve the issue.

dequis:
Sounds like it depends on #442 

Also, trivial conflict in the imports section in ItemTable, just use all from master.

BarelyAPrincess:
@dequis is right, I will remove the [WIP] once a method of removing the door when broken is implemented. ie. #442

I also need to implement redstone support and I can put off actual door removal until then if you'd like to just get this pull request moving. Heck, beds ain't even fully removed yet.

turt2live:
@ChioriGreene psst #442 has been merged and now this isn't mergeable <3

BarelyAPrincess:
It should be fixed now. Just a conflict to the ItemTable as expected. Thanks for getting on this.

gdude2002:
Builds fine here: http://bamboo.gserv.me/browse/GLOW-SRV-PR437-124

turt2live:
Only known issue: When breaking doors in survival mode, it is possible to crash the client.

Tonodus:
This should be wait for and then using #249, making the blockDestroy method obsolete.

Otherwise doors might be half-present, when one half is destroyed by fire, explosion etc.

turt2live:
I disagree. Whichever PR is merged first will not have to fix the bug you've mentioned. No point in holding a PR back for a potential issue.

Tonodus:
I disagree. This isn't a bug, it's just false implemented here (because there's no better option atm). However, seems not to be a big deal for the owner of #249 to fix this anyway...

turt2live:
There's still no real reason to place the entire PR on hold for such a
detail.

sent from mobile

dequis:
> Otherwise doors might be half-present, when one half is destroyed by fire, explosion etc

What exactly is preventing the blockDestroy method from working with fire, explosions, etc?

This already handles destroying the other half block, but it seems to be messing up in some other part (giving the item to the client?)

> However, seems not to be a big deal for the owner of #249 to fix this anyway...

Are you talking about yourself in third person or did you link the wrong PR?

Tonodus:
> What exactly is preventing the blockDestroy method from working with fire, explosions, etc?

They're two different purposes for the methods. This method (blockDestroy) needs to be overwritten, when specific actions must be done based on the player digging the block.
The other method (in the linked PR) is called, whenever the block changes it's type (by breaking by a  player, explosions, fire or even a plugin using setType()) and can therefore not containing information about which player the action does.

> Are you talking about yourself in third person or did you link the wrong PR?

I'm talking about myself in the third-person (in that case).

dequis:
Okay you got me to read the tl;dr discussion you two had in #442.

So, if i'm understanding this right, this is `onBlockChanged` (from #249, ready to pull) vs `blockDestroy` (from #442, merged)

Signatures:

> public void onBlockChanged(GlowBlock block, Material oldType, byte oldData, Material newType, byte data)
> 
> public void blockDestroy(GlowPlayer player, GlowBlock block, BlockFace face, Vector clickedLoc)

And the main difference being discussed here is that `onBlockChanged` takes no GlowPlayer, which makes it better for block changes that aren't from players.

My opinion: Switching from `blockDestroy` to `onBlockChanged` will not change anything regarding the client crash bug, so I agree with "No point in holding the PR back"

Did some testing before writing that: the client crash is indeed because of the (lack of) `getDrops()` implementation. My naive attempts to 'fix' it were returning a stack of `Material.DOOR` and `BlockDropless.EMPTY_STACK`. First one crashed, second one didn't. Turns out [door blocks and door items have different IDs](http://minecraft.gamepedia.com/Door#ID). Annoying, considering that there's a shitton of kinds of doors now. @ChioriGreene, have fun with that.

BarelyAPrincess:
Just my two cents, we need to avoid using methods that don't carry things glowPlayer. I know as a programmer that it can be so frustrating when an api neglects to include such simple information. Forge does a wonderful job with this on a similar method that's called when a player breaks the block but as the mod developer I want to check if the player is in creative, so I can determine if a item drops or not, but I can't since the player is not in the arguments.

The only other option is to reference a source object that caused the change, it could be a player, explosion, entity, etc.

@dequis, I sure I will have fun. It's sad that we have implementation problems because of missing stuff what where. Has anyone considered coming up with a development road map, determining what needs to be implemented at what points and when? We might save some headache verses using all these issues.

dequis:
> Has anyone considered coming up with a development road map, determining what needs to be implemented at what points and when?

That's an extremely vague statement. Feel free to open an issue if you want a discussion like that (we have forums now, but we decided to have development related discussions in github issues instead). But I guess i can answer your question saying that the only roadmap ATM is [stuff like this](https://github.com/GlowstoneMC/Glowstone/milestones/New%20Years%202015) in issue milestones.

> Just my two cents, we need to avoid using methods that don't carry things glowPlayer

It sounds like you're asking for a class type like GlowPlayer that covers not just players but also other entities and other natural sources, so all methods get that. We're kinda tied to the concept of bukkit players, but that might have to change in the future. Open an issue with your proposal?

turt2live:
@ChioriGreene Any update on this? It has been pending changes for a while. Regardless of the lengthy discussion about a method, the client can still be crashed using doors in survival mode (see above).

BarelyAPrincess:
I hope to work on it this weekend, I've been extremely busy lately since I'm getting married in three weeks. My hard drive also went bad, so I had to get all my programming environments reinstalled.

BarelyAPrincess:
I was able to prevent the client crashes by specifying the drops with getDrops(). But without looking deeper I'm sure there is some sort of bug that was directly contributing to the crashes and should probably have a separate issue open. If there was any other problems I needed to look at, let me know.

src/main/java/net/glowstone/net/handler/play/player/DiggingHandler.java
```diff
@@ -9,6 +10,7 @@
 import net.glowstone.entity.GlowPlayer;
 import net.glowstone.net.GlowSession;
 import net.glowstone.net.message.play.player.DiggingMessage;
+
```
turt2live, line 13: Minor note: You have extra whitespace (x2) in this file

turt2live:
@ChioriGreene I've opened a [pull request](https://github.com/ChioriGreene/Glowstone/pull/1) to your branch to include some minor formatting differences and minor code improvements. 

Seeing as they are minor, and that this pull request works, I'm shoving the work on to @SpaceManiac now to pull this with any combination of changes (mine, yours, his, etc). Feel free to merge my PR or not, I've opened it as a convenience for anyone who wants it to be :)

SpaceManiac:
Merged in ce8b93bb15299fbab9e3f79081513eceaf6c0667. Thanks!

BarelyAPrincess:
You know I like how these things will sit and sit and sit then out of the blue tons of things happen and while I'm out on my honeymoon. So I'm not caught up on what just happen until just now when I return. _sigh_

Thanks for the merge... I guess...

Issue #441, opened 2014-10-15
Implement blockDestroy method in the BlockType class


BarelyAPrincess:
I'm creating this issue to present the need for a blockDestroy method in the BlockType class that is called from the DiggingHandler.  This method is needed for the proper removal of doors and beds.  I can make the Pull Request if you think the arguments in the example below are acceptable.

```
public void blockDestroy(GlowPlayer player, GlowBlock block, BlockFace face, Vector blockLoc) {
}
```

SpaceManiac:
Included in 6fba836e68421088be7e960f4e26f0b3b6853a00.

PR #442, opened 2014-10-15
Added method blockDestroy to BlockType class

BarelyAPrincess:
I added this method because it's needed for the proper removal of Doors and Beds. There might also be a need for a separate method for when a block is destroyed by other events like TNT.

---

_Edited by turt2live_

**Related Links:**
- Ticket: #441

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockType.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockType.java
@@ -117,7 +117,7 @@ public class BlockType extends ItemType {
     }
 
     /**
-     * Called when a player attempts to destroys a block.
+     * Called when a player attempts to destroy a block.
      * @param player The player interacting
      * @param block The block the player destroyed
      * @param face The block face
--- a/src/main/java/net/glowstone/block/blocktype/BlockType.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockType.java
@@ -117,6 +117,16 @@ public class BlockType extends ItemType {
     }
 
     /**
+     * Called when a player attempts to destroys a block.
+     * @param player The player interacting
+     * @param block The block the player destroyed
+     * @param face The block face
+     * @param clickedLoc Where the destroy occurred
+     */
+    public void blockDestroy(GlowPlayer player, GlowBlock block, BlockFace face, Vector blockLoc) {
+    }
+
+    /**
      * Called when a player attempts to place a block on an existing block of
      * this type. Used to determine if the placement should occur into the air
      * adjacent to the block (normal behavior), or absorbed into the block
--- a/src/main/java/net/glowstone/block/blocktype/BlockType.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockType.java
@@ -123,7 +123,8 @@ public class BlockType extends ItemType {
      * @param face The block face
      * @param clickedLoc Where the destroy occurred
      */
-    public void blockDestroy(GlowPlayer player, GlowBlock block, BlockFace face, Vector blockLoc) {
+    public void blockDestroy(GlowPlayer player, GlowBlock block, BlockFace face, Vector clickedLoc) {
+        // do nothing
     }
 
     /**
--- a/src/main/java/net/glowstone/net/handler/play/player/DiggingHandler.java
+++ b/src/main/java/net/glowstone/net/handler/play/player/DiggingHandler.java
@@ -1,12 +1,14 @@
 package net.glowstone.net.handler.play.player;
 
-import com.flowpowered.networking.MessageHandler;
 import net.glowstone.EventFactory;
 import net.glowstone.GlowWorld;
 import net.glowstone.block.GlowBlock;
+import net.glowstone.block.ItemTable;
+import net.glowstone.block.blocktype.BlockType;
 import net.glowstone.entity.GlowPlayer;
 import net.glowstone.net.GlowSession;
 import net.glowstone.net.message.play.player.DiggingMessage;
+
 import org.bukkit.Effect;
 import org.bukkit.GameMode;
 import org.bukkit.Material;
@@ -18,6 +20,9 @@ import org.bukkit.event.block.BlockBreakEvent;
 import org.bukkit.event.block.BlockDamageEvent;
 import org.bukkit.event.player.PlayerInteractEvent;
 import org.bukkit.inventory.ItemStack;
+import org.bukkit.util.Vector;
+
+import com.flowpowered.networking.MessageHandler;
 
 public final class DiggingHandler implements MessageHandler<GlowSession, DiggingMessage> {
     @Override
@@ -81,6 +86,10 @@ public final class DiggingHandler implements MessageHandler<GlowSession, Digging
                 BlockPlacementHandler.revert(player, block);
                 return;
             }
+
+            BlockType blockType = ItemTable.instance().getBlock(block.getType());
+            if (blockType != null)
+                blockType.blockDestroy(player, block, face, new Vector(message.getX(), message.getY(), message.getZ()));
 
             // destroy the block
             if (!block.isEmpty() && !block.isLiquid() && player.getGameMode() != GameMode.CREATIVE) {


src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -117,6 +117,16 @@ public boolean blockInteract(GlowPlayer player, GlowBlock block, BlockFace face,
     }
 
     /**
+     * Called when a player attempts to destroys a block.
+     * @param player The player interacting
+     * @param block The block the player destroyed
+     * @param face The block face
+     * @param clickedLoc Where the destroy occurred
+     */
+    public void blockDestroy(GlowPlayer player, GlowBlock block, BlockFace face, Vector blockLoc) {
```
ZephireNZ: Probably should be `Vectory clickedLoc`, to match the others unless I'm missing something.

PaulBGD: FOR VECTORY!

Tonodus:
Related to #249 (duplicate?)

src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -117,6 +117,17 @@ public boolean blockInteract(GlowPlayer player, GlowBlock block, BlockFace face,
     }
 
     /**
+     * Called when a player attempts to destroys a block.
```
turt2live: :( 

1 English instructor has died today.

SpaceManiac:
Is this explicitly needed, or do you think a physics update (from another PR or whatever) on the remaining half of the door/bed to notice the other half is missing would also do the job?

turt2live:
Scheduling a physics update might look a bit strange. As-is it's not very clean when a block is broken in vanilla (half the bed is still there for a moment). This would reduce the lag time between the block breaks and not put extra work into physics as that is bulky and complex enough as-is. 

If a more complicated block shows up in the future I could see that block not using this, but for the time being physics doesn't really have a reason to check for half blocks.

turt2live:
Marking as ready to pull but @SpaceManiac may require some more information from you, @ChioriGreene .

Tonodus:
> This would reduce the lag time between the block breaks and not put extra work into physics as that is bulky and complex enough as-is

IMO the lag you'd expire would be in nanosecond range, blockbreaking is also most part of 249 (as mentioned).

However this should at least be changed so that a plugin or internal event (explosion, fire, ..) setting a block to air with physics enabled (applyPhysics=true) should call this `destroyBlock` method, too... so basically like 249.

BarelyAPrincess:
Just keep in mind that since we are writing for an already established game and most of all the Bukkit API, using a method like this is the already established way of doing it -- I know this from the Forge API.  Doing things different is usually why a mod developer will make a disclaimer like "This mod/plugin will not work on the MCPC+ since it handles threads differently" and since bed and door removal is not expected to be handled by block updates, I don't think it's a good idea.

As I said in my OP, a separate method for TNT, fire, etc will be needed, as this method is for when a physical player tries to remove the block.  Besides, this method will mostly be used for internal events, I don't see it being used by plugins unless they are implementing there own blocks by extending the BlockType class, which at that point, my previous argument would be valid.

Tonodus:
> since bed and door removal is not expected to be handled by block updates, I don't think it's a good idea.

Well, why not? What is 'block updates' for you? For me it's, when a near block changes it's type (i.e. was set to air) notify surrounding blocks.

> As I said in my OP, a separate method for TNT, fire, etc will be needed, as this method is for when a physical player tries to remove the block.

And where's the difference? If a TNT explosion destroys a half door, the other half is expected to be destroyed as well - even if it is not directly by TNT. Same goes for fire.

> keep in mind that since we are writing for an already established game and most of all the Bukkit API, using a method like this is the already established way of doing it

No, it isn't. If a bukkit plugin sets one half of a door to air, the other half disappears as well.

BarelyAPrincess:
> And where's the difference? If a TNT explosion destroys a half door, the other half is expected to be destroyed as well - even if it is not directly by TNT. Same goes for fire.

What's the difference? Umm, a player is actually removing the door, not a fire storm of explosives!

> Well, why not? What is 'block updates' for you? For me it's, when a near block changes it's type (i.e. was set to air) notify surrounding blocks.

I more then know what a block update is, I write a couple mods in Forge.  I know what your trying to get at and using this method has the advantage of being cleaner and straight to the point.  Consider this; if we have the method onNeighborChange() then every single time a block is changed next to the door it's triggered and I get the point that this will be good for when the block under a door is broken but just to check if the top or bottom door is broken will take longer to check (I know. I know. NANOSECONDS).  I think it's far more efficient that when the door is explicitly removed then this method (or blockBlownToSmithereens() or blockBurnedInAnUnmercifulBlazeOfFire() ) is called which at that point we can directly remove the top or bottom half.

> No, it isn't. If a bukkit plugin set one half of a door to air, the other half disappears as well.

I'm pretty certain that setting half a door to air in Bukkit did not trigger the other half to disappear.  In fact, I don't even think it made a block update.  I could be wrong and will probably need to go back to my old Bukkit Plugin (Which had relation to doors ironically.) to find out but I still don't think it changes my previous point.  Besides, if a plugin did explicitly change the top half of a door to air then a method like blockDestroyed() would be called.

Tonodus:
> > And where's the difference? If a TNT explosion destroys a half door, the other half is expected to be destroyed as well - even if it is not directly by TNT. Same goes for fire.
> 
> What's the difference? Umm, a player is actually removing the door, not a fire storm of explosives!

Ofc, but where is the difference in the result? It is the same result, whether a player or a fire or an explosion destroyes a block. So there is no need to have different methods for this cases.

> > Well, why not? What is 'block updates' for you? For me it's, when a near block changes it's type (i.e. was set to air) notify surrounding blocks.*
> 
> [...] Consider this; if we have the method onNeighborChange() then every single time a block is changed next to the door it's triggered and I get the point that this will be good for when the block under a door is broken but just to check if the top or bottom door is broken will take longer to check (I know. I know. NANOSECONDS). I think it's far more efficient that when the door is explicitly removed then this method (or blockBlownToSmithereens() or blockBurnedInAnUnmercifulBlazeOfFire() ) is called which at that point we can directly remove the top or bottom half.

Let's agree that we have different point of views: You prefer the little bit faster method (and I'm not 100% sure it's faster), and I prefer to have one general method that can be easy extended for more cases.

> > No, it isn't. If a bukkit plugin set one half of a door to air, the other half disappears as well.
> 
> I'm pretty certain that setting half a door to air in Bukkit did not trigger the other half to disappear.

It does, tested.

> In fact, I don't even think it made a block update.

It depends, what you call block update. Every time you call Block#setType() a physics update is triggered.

> Besides, if a plugin did explicitly change the top half of a door to air then a method like blockDestroyed() would be called.

That was, what I mentioned, what was missing in your code.

BarelyAPrincess:
I do prefer whatever can get the job done faster because I know in my experience that every nanosecond saved helps.  One difference with it being a player verses a fire or TNT, is item drops.  Should a door or bed drop every time?  If we use blockDestroy, we can make sure the player only receives one door but if we depend on a block update to drop the door then how can we be sure it properly happens or if it should happen at all.  What if a plugin wants to cancel the item drop?  And in this case, if a plugin was to set a block to air, I don't think the door should drop verses a player breaking it.  Again, I feel that using this method will make for cleaner code and allow each block to get straight to the point.

I think we need to wait and see what @SpaceManiac's verdict is.

PaulBGD:
@ChioriGreene http://blog.codinghorror.com/the-sad-tragedy-of-micro-optimization-theater/

Tonodus:
> I do prefer whatever can get the job done faster because I know in my experience that every nanosecond saved helps

Maybe we should write Glowstone in assembly :P seriously, I don't say anything about having a onDestroy method, if it's needed, because we need to know who (which player) breaks a block. But if we don't need to know we should take it in a generic method called for fire etc. too.
**Not only speed matters**. We should write code that is readable, easy to maintain and extend and reusable.

> One difference with it being a player verses a fire or TNT, is item drops. Should a door or bed drop every time? If we use blockDestroy, we can make sure the player only receives one door but if we depend on a block update to drop the door then how can we be sure it properly happens or if it should happen at all. What if a plugin wants to cancel the item drop? And in this case, if a plugin was to set a block to air, I don't think the door should drop verses a player breaking it.

Well, that's not the point. If a player breaks a door block, getDrops() is called and he receives the items. This has **nothing** todo with your onDestroy method.

SpaceManiac:
Merged in 6fba836e68421088be7e960f4e26f0b3b6853a00, with a minor tweak.

PR #452, opened 2014-10-18
Add "server.ip: 0.0.0.0" to default glowstone.yml

bendem:
Because there are quite a few people asking about it and I don't think it costs anything to add it there.

### DIFF
--- a/src/main/resources/defaults/glowstone.yml
+++ b/src/main/resources/defaults/glowstone.yml
@@ -5,6 +5,7 @@
 # For help, join us on IRC: #glowstone @ esper.net
 
 server:
+    ip: 0.0.0.0
     port: 25565
     name: Glowstone Server
     online-mode: true
--- a/src/main/resources/defaults/glowstone.yml
+++ b/src/main/resources/defaults/glowstone.yml
@@ -5,7 +5,7 @@
 # For help, join us on IRC: #glowstone @ esper.net
 
 server:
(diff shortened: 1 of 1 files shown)


SpaceManiac:
I notice a difference in the startup log with this PR applied, likely due to the difference between `ip: ''` and `ip: 0.0.0.0`:

Before: `Binding to address: 0.0.0.0/0.0.0.0:25565...`
After: `Binding to address: /0.0.0.0:25565...`

There appears to be no functional difference, but I was wondering if there's a specific reason you chose `0.0.0.0` instead of the current default blank value which resolves to `InetAddress.anyLocalAddress()`?

dequis:
`ip: ''` would also listen in ipv6, if available. Unless there's a possibility that `InetAddress.anyLocalAddress()` hands out an address that isn't supported by the system - and i hope nobody has fucked up their ipv6 stack that much - that would be preferable.

bendem:
I used 0.0.0.0 because I thought it'd be a nice functional placeholder. I'll change it when I get home.

dequis:
...this seriously doesn't need testing.

SpaceManiac:
Merged in ~~281e3c1~~ 6cafd1d.

bendem:
in 6cafd1d6fd5824034fc16dc229935e8f25eedbf5 actually :wink:

hintss:
@dequis test all the things

PR #457, opened 2014-10-18
Call PlayerMoveEvent

dequis:
Based on JeromSar's code, with several improvements

Particularly, the event isn't called if both locations are equal. This decreases event spam but not as much as craftbukkit does - that would require keeping state in the player of the last sent move event, which would be awkward.

[MiniPython](https://github.com/SpaceManiac/MiniPython) test plugin here: https://gist.github.com/dequis/0c05ebfa93cf9984f877

### DIFF
--- a/src/main/java/net/glowstone/net/handler/play/player/PlayerUpdateHandler.java
+++ b/src/main/java/net/glowstone/net/handler/play/player/PlayerUpdateHandler.java
@@ -14,6 +14,7 @@ public final class PlayerUpdateHandler implements MessageHandler<GlowSession, Pl
     public void handle(GlowSession session, PlayerUpdateMessage message) {
         Location originLoc = session.getPlayer().getLocation();
         Location newLoc = originLoc.clone();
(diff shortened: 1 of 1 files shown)


turt2live:
For those who don't want to compile things or dig for a download link, [slightly different testing materials](https://drive.google.com/folderview?id=0B6JIYSBkLkAnV0luYUloSHFMLW8&usp=sharing&tid=0B6JIYSBkLkAnTTZYd3Z4ZTMzV3c)

turt2live:
@dequis This does not correctly handle the case of a plugin setting the from/to properties. The 'from' property won't do anything until some kind of "You moved too fast" code is implemented, but the 'to' property should move the player there after the event, which is fairly obvious.

SpaceManiac:
Merged in ~~6cafd1d~~ 281e3c1.

bendem:
in 281e3c1b7a49d84e904511b9038ee1adc13e9d44 :wink:

Issue #285, opened 2014-09-14
Red sandstone stairs and slabs are not fully functional


scooterboo:
It should be an easy fix of coping code over form the stairs, slabs, and doubleslabs that work.

Nickforall:
Started working on this, I have to edit Glowkit too

Nickforall:
I'm stuck, I have no idea why it doesn't work, I can't place slabs with half:top.
- I have changed the Glowkit, and added a step2 class, like step and wood_step.
- tried to handle the bytes in the BlockSlab class

but both didn't worked out, I fixed the bug that the slabs couldn't stack and that stairs didn't work.

Tonodus:
I don't understand you well, but the Stair class in bukkit has already methods like get/setFacing and is/setInverted(). You only have to use them and don't have to add a new class.

Nickforall:
I used them, but there is a class that is used for the Material.Step and another one for Material.Wooden_step or something like that. But Material.Step_2 doesn't have one. Those classes have the Inverted methods.

SpaceManiac:
As I understand, fixed in c061facfed82710dff442f96e35138a41048b4c0.

PR #467, opened 2014-10-21
Only drop items if the player mined with the needed tool

Tonodus:
This PR adds checks, whether a Player used the needed tool (e.g. iron pickaxe) to mine a block, otherwise it doesn't drop anything. This checks are in the new class BlockNeedsTool, however every BlockType might override the getDrops() method and made special checks itself, since it now knows the ItemStack a player used.

This PR does not fix the bug that no items are dropped, if the block are insta-breakable, nor does it take enchantments into consideration (e.g fortune, silk touch).

This is related to #98 

---

_Edited by turt2live_

**Related Links:**
- Ticket: #98 
- Ticket: #285


src/main/java/net/glowstone/inventory/ToolType.java
```diff
@@ -0,0 +1,53 @@
...
+    IRON_SWORD(Material.IRON_SWORD, DIAMOND_SWORD),
+    STONE_SWORD(Material.STONE_SWORD, IRON_SWORD),
+    GOLD_SWORD(Material.GOLD_SWORD, STONE_SWORD),
+    SWORD(Material.WOOD_SWORD, GOLD_SWORD);
+
+
+    private final Material bukkitMaterial;
+    private final ToolType better;
+
+    ToolType(Material bukkitMaterial, ToolType better) {
```
turt2live: Should be `private`

dequis:
Simple conflicts in ItemTable:
- `import *` - keep from master
- Added items - keep both sides

Merged branch here: https://github.com/dequis/Glowstone/compare/pr-467

(totally not copypasted from the other two PRs)

turt2live:
Status update: It works, but I do have to manually go through and make sure the correct drops are actually dropped.

turt2live:
@Tonodus After a good 2-3 hours of breaking blocks with @dequis (and some help from @Aaron1011), we've found the following known list of issues with drops. This is no where near a complete list because at about the stopping time we determined that this is really hard and almost impossible to do by hand. Other testing techniques were also considered and none were really adequate in order to have this on the track to being pulled in a reasonable amount of time.

Currently I am pending some kind of feedback from @SpaceManiac on how we should proceed with testing. I'll keep you updated if @SpaceManiac doesn't do so here. I've documented potential testing methods below for transparency. 
- Using some kind of robotic client that breaks blocks in vanilla and Glowstone, comparing drops
  - Involves a lot of time and code. Would also have some logic rules to test.
- Parsing a list of drops from somewhere and writing a unit test
- Using manual block breaking to determine compliance
- (PREFERRED) Looking over the code and manually noting issues, marking this as ready to pull once all known cases are handled. There would be some cases missed and they would be caught by users who open tickets to have them resolved.

I'm more than happy to consider other testing methods from anyone who has ideas. Now, on to the giant list of problems (out of scope issues listed in a list at the end of this post):

**When using your hand to break items...**
- Dark prismarine drops a block
- Prismarine (both variants) drop a block
- Netherrack drops a block
- Dragon eggs drop (should teleport though, so this may be out of scope at the moment)
- Bricks drop a block
- Stone brick (all variants) drop a block
- Red sandstone (all variants) drop a block

**When using a pickaxe (diamond) to break items...**
- Ender chests do not drop the correct items
- Brewing stands crash the client when acquired
- Smooth & chiseled sandstone drop the wrong blocks

**When breaking a block, the following are 'untextured', indicating data value problems**
- Trapped chests
- Ladders
- Wood/Iron trapdoors
- Fence gates (all variants)
- Pumpkins
- Jack-o-lanterns
- Hay bales
- Red sandstone stairs (May be fixed by #503)
- Horizontal logs
- Rails

**When breaking a block, the following do not drop items at all**
- Slime blocks
- TNT
- Torches
- Repeaters

**When breaking a block, the following drop the WRONG items**
- Sea lanterns
- Stained glass panes
- Stained glass blocks
- Stained clay
- Coarse dirt
- Granite (and variants, incl. polished)

**When breaking a block, the following drops cause major issues**
- Lit redstone lamps crash the client
- Doors crash the client (I think this is another issue though)

**The following drop rates are incorrect**
- Saplings
- Seeds (wheat, from tall grass)

I hope these lists are helpful. There are certainly more than that, but those are the confirmed cases in 2-3 hours of testing. 

As for out of scope issues noted while testing this PR... (do **not** fix any of these in this PR, these are listed here for reference **only**).
- Signs are broken beyond belief
- Mob spawners don't drop experience (as with other blocks)
- Walking over repeaters in closed spaces can cause landsharking

dequis:
Some of the things listed in those lists may be considered out of scope too, but they are worth mentioning anyway, just because they appeared during its testing. We could split those to separate issues if needed. The data value problems and client crashing bugs are definitely out of scope, IMO.

turt2live:
From looking at the code, the data values problem is actually a simpler fix than a lot of the issues listed. It is not out of scope.

Edit: with respect to crashing clients.

dequis:
Maybe! Trapped chests are why i'm not sure.

turt2live:
Trapped chests in existing worlds are not problematic (the data value is copied over, hence the untextured texture). It's 'new' trapped chests that cause problems in Glowstone.

dequis:
```
<SpaceManiac> re tool item drops, yes as close as possible is fine, but missing a case
              can just be corrected later
```

Tonodus:
As said, adding/fixing all the stuff would be way too much for a single pr (and this is huge alreay!).

The original idea behind the PR was to **add checks, whether the player used the right tool to mine**, not more nor less. So I decided to add/fix all the blocks that need a special tool (like pickaxe) to drop contents. Every other issue is left for another PR created by me or anyone who's faster.

**Important:** This fixes the same issues as #503 making that PR obsolete. Either that is closed, or I need to remove the fixes this PR makes I assume. It was added for correct dropping btw.

**The list you mentioned for reference:**
~~FIXED~~
_Not handled by this PR_

> When using your hand to break items...
- ~~Dark prismarine drops a block~~
- ~~Prismarine (both variants) drop a block~~
- ~~Netherrack drops a block~~
- _Dragon eggs drop (should teleport though, so this may be out of scope at the moment)_
- ~~Bricks drop a block~~
- ~~Stone brick (all variants) drop a block~~
- ~~Red sandstone (all variants) drop a block~~

> When using a pickaxe (diamond) to break items...
- **Ender chests do not drop the correct items**
  It drops 8xObsidian which is expected, isn't it?
- ~~Brewing stands crash the client when acquired~~
- ~~Smooth & chiseled sandstone drop the wrong blocks~~

> When breaking a block, the following are 'untextured', indicating data value problems
- _Trapped chests_
- _Ladders_
- ~~Wood/Iron trapdoors~~
- _Fence gates (all variants)_
- _Pumpkins_
- _Jack-o-lanterns_
- _Hay bales_
- ~~Red sandstone stairs (May be fixed by #503)~~
- _Horizontal logs_
- _Rails_

> When breaking a block, the following do not drop items at all

**This is caused by insta-break:** #384
- _Slime blocks_
- _TNT_
- _Torches_
- _Repeaters_

> When breaking a block, the following drop the WRONG items
- _Sea lanterns_
- _Stained glass panes_
- _Stained glass blocks_
- ~~Stained clay~~
- _Coarse dirt_
- ~~Granite (and variants, incl. polished)~~

> When breaking a block, the following drops cause major issues
- _Lit redstone lamps crash the client_
- _Doors crash the client (I think this is another issue though)_

> The following drop rates are incorrect
- _Saplings_
- _Seeds (wheat, from tall grass)_

turt2live:
@Tonodus I'd agree with that list. The items not touched by the PR are for another time. As for the ender chests: In testing we did not get any items. @dequis did that test though and should confirm his findings.

dequis:
> This fixes the same issues as #503 making that PR obsolete

RIP my PR (not that I mind)

> **Ender chests do not drop the correct items** It drops 8xObsidian which is expected, isn't it?

Weird. It does work now. Not sure what went wrong before, but blame it on a testing mistake (unless someone reports it again)

turt2live:
For sake of scope, #503 should be pulled on it's own prior to this one so that this PR's scope can be limited. I'm not putting this one on hold just yet, but this one doesn't need to fix the issue described in #503 unless it's a direct requirement for this PRs function (in which case it will have to be decided if we should close #503 or put this one on hold)

dequis:
He said "It was added for correct dropping btw". Just close mine and things will be simpler for everyone, tbh.

turt2live:
@Tonodus @dequis Done. This should also fix the existing issues with stairs here now as well.

Tonodus:
@turt2live All issues with red sandstone stairs / slabs should be fixed.

dequis:
Referencing #285.

Did some more testing, found some more items that drop textureless stuff, but that's unrelated to this PR. We should open an issue for the stuff found here.

Anyway, looks good to me.

turt2live:
Because there are a lot of out of scope issues, #527 has been created to extend #98. Just a heads up.

Some items put into that ticket may be fixed here, but the ticket records "current problems in master".

SpaceManiac:
Merged in c061facfed82710dff442f96e35138a41048b4c0.

PR #469, opened 2014-10-21
Fix ShapedRecipe matching

### LINKED ISSUE #465: CraftingManager matches partially matching small recipes
While trying to craft some wooden sticks I noticed that sometimes I got buttons instead.
- Expected behavior
  RecipeManager will only match completely matching recipes
- Current behavior
  RecipeManager will match partially matching small recipes
- How to reproduce
  1. Place 2 wood planks diagonally in the crafting grid
  2. Click the empty result slot (This will force the client inventory view to sync with the server inventory view)
  3. Obtain button from result slot

ghost:
This pull request includes improved ShapedRecipe-items matching and introduces an additional check for it. Fixes: #465. 

**Modifications**
By moving the `position` label to the column-start loop, possible recipes aren't skipped due to only being available in at a higher column number, when a lower column number check fails. In addition to that the `ingredientChar` is defined as the appropriate char in the current row of the shape, or if it isn't available in there, a space. The additional check simply examines the existence of items outside the validated recipe part of the given items.

**Result**
More reliable crafting of shaped recipes.

### DIFF
--- a/src/main/java/net/glowstone/inventory/CraftingManager.java
+++ b/src/main/java/net/glowstone/inventory/CraftingManager.java
@@ -138,14 +138,14 @@ public final class CraftingManager implements Iterable<Recipe> {
             if (rows == 0 || cols == 0) continue;
 
             // outer loop: try at each possible starting position
-            position:
             for (int rStart = 0; rStart <= size - rows; ++rStart) {
+                position:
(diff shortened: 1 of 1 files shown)


SpaceManiac:
Merged in 400ae0c5495c3bf6eb1f2e5d9a5430c4d9f28bca.

PR #472, opened 2014-10-22
Fix Hoe interaction with coarse dirt and podzol

jimmikaelkael:
I introduced a bug related to farming with Hoe interaction: the block data isn't checked and so it's possible to turn podzol and coarse dirt into soil.
The expected behavior in vanilla 1.8 is that podzol is not interactable with Hoe, and coarse dirt turns into normal dirt.

This commit fixes the problem.


SpaceManiac:
Merged in acc5ed2000847d4145dfef8174aadd658acf8e90.

Issue #154, opened 2014-09-04
EncryptionKeyResponseHandler error on connection cancel


SpaceManiac:
Clicked 'cancel' while logging in, produced this error (trying to manipulate pipeline of closed connection?):

```
15:35:33 [SEVERE] Error while handling net.glowstone.net.message.login.EncryptionKeyResponseMessage@5ad59375 (handler: EncryptionKeyResponseHandler)
java.util.NoSuchElementException: encryption
    at io.netty.channel.DefaultChannelPipeline.getContextOrDie(DefaultChannelPipeline.java:907)
    at io.netty.channel.DefaultChannelPipeline.replace(DefaultChannelPipeline.java:377)
    at net.glowstone.net.GlowSession.updatePipeline(GlowSession.java:404)
    at net.glowstone.net.GlowSession.enableEncryption(GlowSession.java:390)
    at net.glowstone.net.handler.login.EncryptionKeyResponseHandler.handle(EncryptionKeyResponseHandler.java:78)
    at net.glowstone.net.handler.login.EncryptionKeyResponseHandler.handle(EncryptionKeyResponseHandler.java:33)
    at com.flowpowered.networking.session.BasicSession.handleMessage(BasicSession.java:82)
    at com.flowpowered.networking.session.BasicSession.messageReceived(BasicSession.java:141)
    at net.glowstone.net.GlowSession.pulse(GlowSession.java:356)
    at net.glowstone.net.SessionRegistry.pulse(SessionRegistry.java:23)
    at net.glowstone.scheduler.GlowScheduler.pulse(GlowScheduler.java:172)
    at net.glowstone.scheduler.GlowScheduler.access$100(GlowScheduler.java:24)
    at net.glowstone.scheduler.GlowScheduler$2.run(GlowScheduler.java:109)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
    at java.util.concurrent.FutureTask$Sync.innerRunAndReset(FutureTask.java:351)
    at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:178)
    at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:178)
    at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
[...]

PR #473, opened 2014-10-22
Fix race conditions

### LINKED ISSUE #421: Rare exception when 40+ clients disconnect in the same tick
```
2014/10/12 14:29:10 [SEVERE] Error occurred while pulsing world world
java.util.ConcurrentModificationException
    at java.util.HashMap$HashIterator.nextEntry(HashMap.java:922)
    at java.util.HashMap$ValueIterator.next(HashMap.java:950)
    at net.glowstone.entity.GlowPlayer.pulse(GlowPlayer.java:416)
    at net.glowstone.GlowWorld.pulse(GlowWorld.java:371)
    at net.glowstone.scheduler.WorldScheduler$WorldThread.run(WorldScheduler.java:48)
```

No other information available.

coelho:
There are actually a lot of race conditions in Glowstone's network implementation. This PR fixes two notable ones that were reported.
#421

This is caused by the player entity being removed from an asynchronous thread while the synchronous thread is iterating over the map of entities.
This is resolved by removing the player entity from the synchronous thread.
#154

This is actually a very rare race condition, but it is nice to see that it was reproduced. This is where the channel is disconnected asynchronously at the exact same time that the GlowSession.pulse(); synchronous method is called. Because the EncryptionKeyResponse is handled on the synchronous thread, we would then handle it, and since we are not in the PlayProtocol state yet, we wouldn't acknowledge that the session has disconnected.

Due to the way that we fix #421 adding a "disconnected" variable instead of relying on it being in the PlayProtocol to be able to notice if it's disconnected, this is naively fixed or the window is actually narrowed for this issue to be a lot more slim, however it is not a proper fix, and there is still definitely a silent race condition.

The real solution to this issue is to handle the EncryptionKeyResponse asynchronously.

### DIFF
--- a/src/main/java/net/glowstone/net/GlowNetworkServer.java
+++ b/src/main/java/net/glowstone/net/GlowNetworkServer.java
@@ -39,7 +39,7 @@ public final class GlowNetworkServer implements ConnectionManager {
 
     @Override
     public Session newSession(Channel c) {
-        GlowSession session = new GlowSession(server, c);
+        GlowSession session = new GlowSession(server, c, this);
         server.getSessionRegistry().add(session);
         return session;
     }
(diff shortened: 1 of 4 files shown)


turt2live:

PR #267, opened 2014-09-12
Skulls

ZephireNZ:
This PR adds in Mob Skulls, Named Player Skulls, and the Skull item, as well as all of the necessary Bukkit APIs.

Most seems to be working well. Player can place skulls on the floor/walls, rotation works on the floor, and named player heads have their skins downloaded. All of this is saved and loaded properly. Easiest way to test names is with a heads plugin, such as [PlayerHeads](http://ci.md-5.net/job/PlayerHeads/).

Thanks to @Aaron1011 and @jkcclemens for their help with this.

PR #477, opened 2014-10-23
Add flower pot implementation

ghost:
For this pull request is this pull request for Glowkit required: Glowkit#48. This basically adds all the necessary implementation for it.

**Modifications**
Add the implementation, as block state, for `org.bukkit.block.FlowerPot` introduced in the previously mentioned pull request. This is the only class that distinguishes the pre-1.7 way of handling flower pot contents (block data) and the post-1.7 way (tile entities). This implementation is also now registered to the item table. Apart from that a flower pot tile entity and block type implementation are added.

**Result**
Proper past, present and future proof handling of flower pots.

---

_Edited by turt2live_

**Related Links:**
- Glowkit: GlowstoneMC/Glowkit#48
- Ticket: #325

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockFlowerPot.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockFlowerPot.java
@@ -59,8 +59,7 @@ public class BlockFlowerPot extends BlockType {
             // Only change contents if there is none and if the held item is valid pot contents.
             if (pot.getContents() == null && heldItem != null && isValidContents(heldItem.getData())) {
                 pot.setContents(heldItem.getData().clone()); // Null-check in isValidContents.
-                pot.update();
-                return true;
+                return pot.update();
             }
         }
         return false;
--- a/src/main/java/net/glowstone/block/blocktype/BlockFlowerPot.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockFlowerPot.java
@@ -25,7 +25,7 @@ import net.glowstone.entity.GlowPlayer;
 public class BlockFlowerPot extends BlockType {
 
     @Override
-    public Collection<ItemStack> getDrops(GlowBlock block) {
+    public Collection<ItemStack> getDrops(GlowBlock block, ItemStack tool) {
         List<ItemStack> drops = Arrays.asList(new ItemStack(Material.FLOWER_POT));
         GlowBlockState state = block.getState();
 
--- a/src/main/java/net/glowstone/block/blocktype/BlockFlowerPot.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockFlowerPot.java
@@ -1,0 +1,66 @@
+package net.glowstone.block.blocktype;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+import org.bukkit.Material;
+import org.bukkit.block.BlockFace;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.material.FlowerPot;
+import org.bukkit.material.MaterialData;
+import org.bukkit.util.Vector;
+
+import net.glowstone.GlowChunk;
+import net.glowstone.block.GlowBlock;
+import net.glowstone.block.GlowBlockState;
+import net.glowstone.block.entity.TEFlowerPot;
+import net.glowstone.block.entity.TileEntity;
+import net.glowstone.block.state.GlowFlowerPot;
+import net.glowstone.entity.GlowPlayer;
+
+public class BlockFlowerPot extends BlockType {
+
+    @Override
+    public Collection<ItemStack> getDrops(GlowBlock block) {
+        List<ItemStack> drops = Arrays.asList(new ItemStack(Material.FLOWER_POT));
+        GlowBlockState state = block.getState();
+
+        if (state instanceof GlowFlowerPot) {
+            MaterialData contents = ((GlowFlowerPot) state).getContents();
+
+            if (contents != null) {
+                drops.add(new ItemStack(contents.getItemType(), 1, contents.getData()));
+            }
+        }
+        return Collections.unmodifiableList(drops);
+    }
+
+    @Override
+    public TileEntity createTileEntity(GlowChunk chunk, int cx, int cy, int cz) {
+        return new TEFlowerPot(chunk.getBlock(cx, cy, cz));
+    }
+
+    @Override
+    public boolean blockInteract(GlowPlayer player, GlowBlock block, BlockFace face, Vector clickedLoc) {
+        GlowBlockState state = block.getState();
+        MaterialData data = state.getData();
+
+        if (!(data instanceof FlowerPot)) {
+            warnMaterialData(FlowerPot.class, data);
+            return false;
+        }
+        if (state instanceof GlowFlowerPot) {
+            GlowFlowerPot pot = (GlowFlowerPot) state;
+            // Only change contents if there is none.
+            if (pot.getContents() == null) {
+                // Todo: check if the item is valid (null-check too) as flower pot contents.
+                pot.setContents(player.getItemInHand().getData().clone());
+                pot.update();
+                return true;
+            }
+        }
+        return false;
+    }
+}
--- a/src/main/java/net/glowstone/block/blocktype/BlockFlowerPot.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockFlowerPot.java
@@ -5,10 +5,12 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 
+import org.bukkit.GrassSpecies;
 import org.bukkit.Material;
 import org.bukkit.block.BlockFace;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.material.FlowerPot;
+import org.bukkit.material.LongGrass;
 import org.bukkit.material.MaterialData;
 import org.bukkit.util.Vector;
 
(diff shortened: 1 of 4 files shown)


src/main/java/net/glowstone/block/blocktype/BlockFlowerPot.java
```diff
@@ -0,0 +1,66 @@
...
+            warnMaterialData(FlowerPot.class, data);
+            return false;
+        }
+        if (state instanceof GlowFlowerPot) {
+            GlowFlowerPot pot = (GlowFlowerPot) state;
+            // Only change contents if there is none.
+            if (pot.getContents() == null) {
+                // Todo: check if the item is valid (null-check too) as flower pot contents.
+                pot.setContents(player.getItemInHand().getData().clone());
+                pot.update();
```
Aaron1011: The `update` method doesn't appear to be overriden in`TEFlowerPot`, so this currently does nothing.

ghost: The invocation of `GlowFlowerPot#update()` calls `TileEntity#updateInRange()` if it's a post-1.7 pot, which in turn invokes `TileEntity#update(GlowPlayer)`. This is indeed not overridden by `TEFlowerPot`, but neither is it in other tile entity classes, except for `TESign`. Furthermore there isn't a method to send tile entity updates, so I thought it would fall out of the scope of this pull request to make one.

ZephireNZ: An implementation of the TE update packets and their use in a TE is in #267.

The reason why none of the other TE's override `TileEntity#update(GlowPlayer)` is because there is no visible changes upon NBT change. Whereas for items such as signs, banners, and pots, a change in NBT likely means a visible change - so that NBT should be sent to the player.

ghost: I saw it, but didn't find it fitting to include in this pull request. I personally think it would be much more practical to write a separate pull request for updating relevant tile entities to easily maintain consistency. And yes, you're correct, it is for visual updates mostly, yet `TEMobSpawner` among one of the tile entities that require visual updates, hasn't overridden `TileEntity#update(GlowPlayer)`...

turt2live:
`GLOWKIT: 48`

turt2live:
@alexspieksma Your PR has a problem:

1) Place a flower pot on the ground
2) Attempt to place a sand block on top of the empty pot
- It will show for a moment but disappear

3) Place a sapling in the pot
- A sapling will appear above and inside the pot

ghost:
@turt2live The issue listed under `2)` is due to the fact that there's no filter for interacting with the pot with invalid items. This is commented on, on line _58_ in `BlockFlowerPot.java`. What actually happens is that the client thinks the sand block is placed on top of the pot, while the server handles it as putting the sand block in the pot. I decided not to implement this check, since these kind of checks for e.g. inventories, such as brewer inventories, are also not included yet.

The problem under `3)` is a little bit more interesting. In this case the pot is already filled, since the sand block has been put in and thus `false` is returned in `BlockFlowerPot#blockInteract(GlowPlayer, GlowBlock, BlockFace, Vector)`. As a result the sapling is placed, by the server, on the side of the pot you clicked on, because apart from that, Glowstone currently has **no (or nearly no) implementation for `BlockType#canPlaceAt(GlowBlock, BlockFace)` and relies heavily on the client instead**. The sappling also appears in the pot, because `TileEntity#update(GlowPlayer)` hasn't been overridden as explained in the previous comments. Causing the client to not know there's a sand block in the pot and rendering a sapling in it.

turt2live:
I do not think that this should be accepted without proper implementation then. This does not mean going around fixing all of the existing logical components, but it does mean that flower pots should function correctly. If the resulting system can be potentially carried over to other logical components in the game then the fix should be designed with that in mind.

ghost:
@turt2live Alright. I'll include checks for valid pot contents and override `TileEntity#update(GlowPlayer)` in `TEFlowerPot` after https://github.com/GlowstoneMC/Glowstone/pull/267 has been pulled, because it contains the necessary components for sending tile entity changes to clients. However, being able to place plants on the sides of pots seems unrelated to this pull request and more like something that should be fixed separately. Signs, for example, can also be placed on, in vanilla, invalid blocks. Although adding the aforementioned features may "fix" this placement issue. What do you think?

turt2live:
The issue of placing plants on the sides of the pot is not the concern as it is an unintended side effect problem. Resolving the concern that flower pots can contain invalid contents will more than likely resolve other concerns relating to invalid block placement.

`PRHOLD: Waiting for #267 - Requires entity change code`

turt2live:
Hold cleared: Skulls have been pulled. Pending further review.

ghost:
@turt2live After implementing `TileEntity#update(GlowPlayer)` in `TEFlowerPot` and some testing, it seems like the client handles all the valid contents checks before sending a packet to the server. So is including checks for valid pot contents still necessary in `BlockFlowerPot#blockInteract(GlowPlayer, GlowBlock, BlockFace, Vector)`, in this pull request?

turt2live:
Yes, because the client may not be the Mojang Vanilla client. In general, assume that the client is not following expected behaviour and bullet-proof everything that the client sends. I know Glowstone doesn't do much of this as-is, but going forward Glowstone will be doing so. Existing code in Glowstone will be improved to match.

ghost:
@turt2live Should be complete now!

turt2live:
This has several issues with loading flower pots created in other worlds. 

**1.8 test**

_Vanilla_
vanilla1

_This PR_
pr1

**1.7.10 test**

_Vanilla_
vanilla1

_This PR_
pr2

Please fix!

turt2live:
@alexspieksma Your PR does not compile

ghost:
@turt2live I made a few modifications and tested them by importing a vanilla 1.7.10 and 1.8.1-pre3 flatworld, containing all valid pot contents in flowerpots, into Glowstone. Everything worked as far as I could tell, although there seems to be one problem. All flowerpots are visually empty - only visually because pointing at the pots shows the correct contents in the f3-menu - and after placing a block near them, the contents appears. I'm not sure what causes this oddity though.

By the way, the current code converts pots that previously used material IDs in their NBT data to pots that use material ID names. Vanilla does this as well.

turt2live:
Sounds like you need to send block updates.

ghost:
@turt2live Actually, it's a problem with all tile entities using messages for visual updates: see https://github.com/GlowstoneMC/Glowstone/issues/547.

src/main/java/net/glowstone/block/blocktype/BlockFlowerPot.java
```diff
@@ -0,0 +1,90 @@
...
+            warnMaterialData(FlowerPot.class, data);
+            return false;
+        }
+        if (state instanceof GlowFlowerPot) {
+            GlowFlowerPot pot = (GlowFlowerPot) state;
+            ItemStack heldItem = player.getItemInHand();
+            // Only change contents if there is none and if the held item is valid pot contents.
+            if (pot.getContents() == null && heldItem != null && isValidContents(heldItem.getData())) {
+                pot.setContents(heldItem.getData().clone()); // Null-check in isValidContents.
+                pot.update();
```
turt2live: You ignore the return value and assume it works here. Any reason why?

src/main/java/net/glowstone/block/state/GlowFlowerPot.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public void setContents(MaterialData contents) {
+        this.contents = contents;
+    }
+
+    @Override
+    public boolean update(boolean force, boolean applyPhysics) {
+        // Pre-1.7 uses block data.
```
turt2live, line 45: Shouldn't a conversion be done well before this point?

ghost, line 45: Huh? I'm not sure what conversion you mean, but if you're talking about storing whether the pot is pre- or post-1.7: knowing whether it is, between state creation and the invocation of this method, seems useless to me.

src/main/java/net/glowstone/block/blocktype/BlockFlowerPot.java
ghost: Currently it will only fail if the state stores a different type value than the block. Chances are 0 this will ever happen at this spot. Returning the return value of `GlowBlockState#update()` doesn't look like it will hurt though. Although that method should implicitly rollback persistent block changes if it fails after setting the block's data. If it doesn't, the item in the player's hand should be used (thus returning _true_ instead of _false_), although I might have been a bit preemptive there.

turt2live: This should definitely use the return value

src/main/java/net/glowstone/block/state/GlowFlowerPot.java
turt2live, line 45: From what I can tell though, the pot is saved in the same format it was loaded in. This is false behaviour and should be corrected. 

If it's an old version then it should be converted upon being loaded. After loading, only the newest version will be used, including for saving.

ghost, line 45: False behaviour? Pre-1.7 pots are saved in vanilla single player as pre-1.7 pots in higher versions of the game: they aren't converted in any way. This pull request does it exactly like that.

turt2live, line 45: Do both flower pot API endpoints work fully then? Both the material data and block state should function the same (or to the best of their ability). 

I apologize for assuming the Mojang actually designed their game correctly though. Generally a conversion would be done. I'll have to talk to others to see if it's worth ignoring vanilla in this case.

ghost, line 45: If it is a pre-1.7 pot, both material data and block state will return the same. However, if it is a post-1.7 pot, only block state will return the correct data, since it isn't possible for the material data to know about the tile entity. So additionally, converting the pot to post-1.7 would make the material data of it utterly useless.
As a reminder: the linked Glowkit pull request deprecates the material data set and get contents functions.

turt2live:
@alexspieksma any update on this? It's been pending changes for a while

ghost:
@turt2live I fixed the piece of code you previously commented on. However, flower pots will still seem buggy (being invisible on join from time to time) due to https://github.com/GlowstoneMC/Glowstone/issues/547 still requiring a fix.

turt2live:
@alexspieksma This pull request [failed to compile](http://home.turt2live.com:55/build/log?id=232&html#L1119-L1123), please resolve the issue as soon as possible.

turt2live:
~~@alexspieksma This still doesn't function for a simple case of saplings (didn't test past that). The saplings appear on top of the pots and inside the pots, but logging out and back in proves the pots to be empty (checked via a plugin as well).~~

Edit: I didn't even test to with this PR pulled...

turt2live:
@alexspieksma disregard my last comment, I never actually tested the PR by accident ._.

However, 1.7.10 -> 1.8 conversion failed as shown below:

1.7.10:


1.8 (Glowstone):

ghost:
@turt2live Try placing a block next to one of the flowerpots. When I encountered the same issue as displayed in your screenshots this resulted in the pots updating client-sided and showing the correct contents. I'm confident this has to do with https://github.com/GlowstoneMC/Glowstone/issues/547 - in both cases the client does know the correct contents (check the f3 menu while pointing at a pot), but somehow refuses to display it visually.

turt2live:
I can't consider this complete unless it works as expected though. If that means it needs another implementation to be present, then this can be put on hold.

turt2live:
@alexspieksma any comment on my most recent response?

ghost:
@turt2live If you regard this as not working as expected because visually changeable tile entities don't appear properly on join, then putting this pull request on hold until https://github.com/GlowstoneMC/Glowstone/issues/547 is fixed is the only viable option.

turt2live:
The issue I'm having is that I can't reproduce the tile entities issue outside of this PR at all. I've tried the steps on the mentioned ticket with no success. When I have a lot more time I will be testing this and the mentioned issue even further (although I've already sunk 3x the normal amount of testing into this PR). Steps to reproduce the issue 100% of the time (or more than 0%, really) would be appreciated.

PR #479, opened 2014-10-24
[WIP] Add vanilla-like dungeons generation

jimmikaelkael:
This will add dungeons generation.


dequis:
In case someone else is wondering the same thing as me - this actually extends/modifies the existing DungeonPopulator to match vanilla more closely.

jimmikaelkael:
I'm closing this PR in favor of #524.

Issue #482, opened 2014-10-24
Some control over messages should be present


Cadiducho:
Add more configurable messages, like the reason of kick when the server is full, when the server is whitelisted (and you aren't) and some more like that

turt2live:
PR: https://github.com/GlowstoneMC/Glowstone/pull/483

turt2live:
I'm closing this in favour of #544 which has a more broad scope.

PR #483, opened 2014-10-24
Add more configurable messages

Cadiducho:
Added more configurable messages, such as kick reason when you aren't in the whitelist or the server is full

---

_Edited by turt2live_

**Related Links:**
- Ticket: #482


turt2live:
I'm not sure this is needed. Do you have a use case that demonstrates how this is required? A plugin can easily override these...

Cadiducho:
For example, a 'default' translate for another language. Yes, this can be override these; but this could well be part of Glowstone configuration, as already include Spigot

turt2live:
I'm going to assign this to @SpaceManiac for further review as I am unsure as if this is even needed. I personally do not think so, but @SpaceManiac may have other ideas. Community feedback is welcome.

Indicardo:
it can be usable :D

bendem:
There are a bunch of trailing whitespaces in that diff.

SpaceManiac:
It seems to me like most of these can be covered with plugins. A couple of the message changes you make in the defaults (notably with outdated client/server) are worth adding on their own, but I'm not sure this is the right way to go about configuring them. I'm not sure whether it would make more sense to add a more general message-configuration system (a la `.lang` files?) or just relegate as much as possible to the plugin realm.

Cadiducho:
Is it possible to create a messages.yml or messages.lang? Yes it's possible, only respected the organization all together in the file 'glowstone.yml' and separate subcategories depending on the structure of YAML 
I don't think that all of these changes can be implemented by plugins, and I don't think that one plugin is necessary for such useful changes (especially for non-English speaking people).

turt2live:
I think going with a lang file would be a better way forward for this. As for what messages are needed in the lang file, I'm not sure. I don't imagine that we need an extensive list of messages to be in there (as there are way too many in the server code), but some options would be nice.

turt2live:
@Cadiducho This pull request is no longer mergeable. Please resolve the issue as soon as possible, thanks!

Cadiducho:
Sure. I think that some files like "messages-en.lang", "messages-fr.lang" could be great... Right? 
Simply you select 'en', 'fr' (etc) in 'glowstone.yml'. Most clear.

Tonodus:
> Simply you select 'en', 'fr' (etc) in 'glowstone.yml'.

Would be great if the messages would be 'translated' automatically based on the clients settings; might however be better suited as a plugin.

turt2live:
@Cadiducho Your pull request has been recorded as not mergeable for 7 days now. Please resolve the conflicts as soon as possible for review to continue, thanks.

dequis:
Merged branch: https://github.com/dequis/Glowstone/compare/pr-483

The last merge in here seems to have rebased things backwards. Did a reset --hard 3caef8dc294fbbbb8abd0fcd5aa09f36a33431a8.

turt2live:
@Cadiducho This pull request is all kinds of weird right now. Please respond to this message so I can rest assured that you aren't dead or missing in action.

I also don't see why these can't be covered by plugins. I know the point was brought up before, but I can't see a definitive reason to have the extensive language support proposed in this request. Please address in a comment why this proposed system would be more useful than a plugin. I understand many servers are hosted outside of English speaking countries, but our website, documentation, code, etc are all English. The implication of having server-side support for different languages suggests that we actually support the languages project-wide, which we don't (currently).

turt2live:
Suggesting closure to @SpaceManiac for the following reasons:
- Unclear intentions (see recent comments)
- Strange state for merge
- Inactive author

SpaceManiac:
Closing for reasons previously described.

PR #431, opened 2014-10-14
Add basic random blocks tick

jimmikaelkael:
This PR needs #432 to be pulled first.
#### Progress
- [x] modify the block coordinates random to use only one nextInt() call
- [x] restrict the chunk tick list to a list of chunks around each player
- [x] improve block growth to match vanilla more closely in BlockCrops, BlockNetherWart & BlockStem (chances, other blocks proximity, etc...)
- [x] provide a test plugin for BlockSpreadEvent, BlockFadeEvent, BlockGrowEvent and StructureGrowEvent.
#### Test Plugin:
- jar: https://drive.google.com/file/d/0B0gnjB7PN1h3blZuZlgzOGV1X0k/view?usp=sharing
- source available here: https://gist.github.com/jimmikaelkael/e5812bcf4c7b098a6d6a

---

_Edited by turt2live_

**Related Links:**
- Ticket: #40 
- Waiting on #432 (or similar)

PR #497, opened 2014-10-31
Add better thunder effect

jimmikaelkael:
This PR adds sounds to thunder and better randoms in lightning.

---

_Edited by turt2live_

**Related Links:**
- No ticket (none needed)

### DIFF
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -188,19 +188,9 @@ public final class GlowWorld implements World {
     private int thunderingTicks = 0;
 
     /**
-     * The rain density on the previous world tick.
-     */
-    private float previousRainDensity = 0;
-
-    /**
      * The rain density on the current world tick.
      */
     private float currentRainDensity = 0;
-
-    /**
-     * The sky darkness on the previous world tick.
-     */
-    private float previousSkyDarkness = 0;
 
     /**
      * The sky darkness on the current world tick.
@@ -1188,19 +1178,13 @@ public final class GlowWorld implements World {
             previouslyRaining = currentlyRaining;
         }
 
-        float rainModifier, skyDarknessModifier;
-        if (currentlyRaining) {
-            rainModifier = .01F;
-        } else {
-            rainModifier = -.01F;
-        }
-        if (currentlyThundering) {
-            skyDarknessModifier = .01F;
-        } else {
-            skyDarknessModifier = -.01F;
-        }
+        float previousRainDensity = currentRainDensity;
+        float previousSkyDarkness = currentSkyDarkness;
 
-        currentRainDensity = (float) Math.max(0, Math.min(1, previousRainDensity + rainModifier));
+        final float rainDensityModifier = currentlyRaining ? .01F : -.01F;
+        final float skyDarknessModifier = currentlyThundering ? .01F : -.01F;
+
+        currentRainDensity = (float) Math.max(0, Math.min(1, previousRainDensity + rainDensityModifier));
         currentSkyDarkness = (float) Math.max(0, Math.min(1, previousSkyDarkness + skyDarknessModifier));
 
         if (previousRainDensity != currentRainDensity) {
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -1287,7 +1287,7 @@ public final class GlowWorld implements World {
     public void playSound(Location location, Sound sound, float volume, float pitch) {
         if (location == null || sound == null) return;
 
-        final double radiusSquared = Math.pow(volume > 1 ? volume * 16 : 16, 2);
+        final double radiusSquared = Math.pow(volume * 16, 2);
         for (Player player : getRawPlayers()) {
             if (player.getLocation().distanceSquared(location) <= radiusSquared) {
                 player.playSound(location, sound, volume, pitch);
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -178,11 +178,6 @@ public final class GlowWorld implements World {
     private int rainingTicks = 0;
 
     /**
-     * Whether it was thundering on this world in the previous tick.
-     */
-    private boolean previouslyThundering = false;
-
-    /**
      * Whether it is currently thundering on this world.
      */
     private boolean currentlyThundering = false;
@@ -1176,12 +1171,11 @@ public final class GlowWorld implements World {
 
     private void updateWeather() {
 
-        if (previouslyRaining != currentlyRaining || previouslyThundering != currentlyThundering) {
+        if (previouslyRaining != currentlyRaining) {
             for (GlowPlayer player : getRawPlayers()) {
                 player.sendWeather();
             }
             previouslyRaining = currentlyRaining;
-            previouslyThundering = currentlyThundering;
         }
 
         float previousRainDensity = currentRainDensity;
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -178,6 +178,11 @@ public final class GlowWorld implements World {
     private int rainingTicks = 0;
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1218,7 +1218,7 @@ public void playEffectExceptTo(Location location, Effect effect, int data, int r
     public void playSound(Location location, Sound sound, float volume, float pitch) {
         if (location == null || sound == null) return;
 
-        final double radiusSquared = Math.pow(Math.min(volume * 16, 16), 2);
+        final double radiusSquared = Math.pow(volume > 1 ? volume * 16 : 16, 2);
```
turt2live: What was the issue?

src/main/java/net/glowstone/entity/GlowLightningStrike.java
```diff
@@ -25,10 +27,12 @@
      */
     private final int ticksToLive;
 
+    private final Random random = new Random();
+
     public GlowLightningStrike(Location location, boolean effect) {
         super(location);
         this.effect = effect;
-        this.ticksToLive = 30;
+        this.ticksToLive = random.nextInt(3) + 1;
```
turt2live: This is a significant change. What is the reason?

jimmikaelkael: This is the goal of this PR... Randomize the lightning living time.

The random object could be passed in constructor.

src/main/java/net/glowstone/GlowWorld.java
jimmikaelkael: Without this change you'll only hear lightning if it strike very near to you but you'll see the flash, while the thunder is something you hear from far ditance in vanilla.

The current implementation will not play sounds to players if their squared distance from the sound location is more than 256 which is a very small number.
Say you're at location 0, 64, 0, lightning strike at 70, 64, 50 which is something not really far away, the squared distance is 70_70 + 0_0 + 50*50 = 7400.
Even worst say it strikes at 12, 64, 12, this time you'll really see it ^^, but your squared distance is already 288 so with the current implementation you'll almost never hear the sound...

What I'm doing match more closely how vanilla is treating sound effects radius.

One a side note, actually lightning can strike on every loaded chunks while it should strike in active players chunks (ie after #431 is pulled).

jimmikaelkael:
@turt2live I added smooth transitions between weather changes concerning the rain density and sky darkness. During thunder the sky becomes even darker.

src/main/java/net/glowstone/GlowWorld.java
turt2live: This is still incorrect though. A sound played at 0.5f volume should be heard within 8 blocks of the source, not 16.

jimmikaelkael: Should be good now.

src/main/java/net/glowstone/entity/GlowLightningStrike.java
```diff
@@ -25,10 +27,13 @@
...
     private final int ticksToLive;
 
-    public GlowLightningStrike(Location location, boolean effect) {
+    private final Random random;
+
+    public GlowLightningStrike(Random random, Location location, boolean effect) {
         super(location);
         this.effect = effect;
-        this.ticksToLive = 30;
+        this.ticksToLive = random.nextInt(3) + 1;
```
turt2live: I'm still looking for verification that this is the correct change. I can't find any information on what this number is supposed to be

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -1681,6 +1688,14 @@ public void sendWeather() {
         session.send(new StateChangeMessage(stormy ? 2 : 1, 0));
     }
 
+    public void sendRainDensity() {
```
turt2live, line 1691: Why can't these be sent in `sendWeather`?

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -183,6 +188,26 @@ protected String disambiguate(World subject, String metadataKey) {
     private int thunderingTicks = 0;
 
     /**
+     * The rain density on the previous world tick.
+     */
+    private float previousRainDensity = 0;
```
SpaceManiac: It seems to me like these `previous*` fields aren't needed and can just be initialized in the `updateWeather()` method before the `current*` fields are changed.

src/main/java/net/glowstone/entity/GlowPlayer.java
jimmikaelkael, line 1691: @turt2live Because it's not the same: SendWeather send the StateChangeMessage packet once with a value of 1 or 2 to stop or start raining.

During a transition, the rain density and sky darkness changes smoothly with a step of 0.01F on each tick: in this case the StateChangeMessage packet is sent more frequently but weather doesn't change (start/stop raining) so there's no need to resend it.

src/main/java/net/glowstone/entity/GlowLightningStrike.java
jimmikaelkael: This is the living duration. the lighting itself can vary the ticks lived. Without doing this you would have longer and less random flashes instead of those blinking flashes.

SpaceManiac: Could you do testing to verify this actually changes anything? In my experience, the lightning bolt will show for a short, random time on the client regardless of what value is filled in here.

src/main/java/net/glowstone/GlowWorld.java
jimmikaelkael: This is true, thank you. It seems I only need to keep previouslyRaining to track weather changes by command or plugins calling SetStorm() directly. Will change this asap.

src/main/java/net/glowstone/entity/GlowLightningStrike.java
jimmikaelkael: Yep I already tested it. As I mentionned above, I tried without impacting on the time to live: once I added sound the lightning flash became longer thant it should be and it was really not realistic.

I'll verify it again to be sure.

src/main/java/net/glowstone/entity/GlowPlayer.java
turt2live, line 1691: Do you really need 3 methods though? 

From the looks of it you can reduce sky and rain to one method and have that called from sendWeather. This reduces the number of methods called elsewhere as well.

jimmikaelkael, line 1691: SendWeather method shouldn't send these packets: I you set the weather for a specific player, you don't want to send them densities unless you can track these values for each players, but anyway Bukkit API doesn't handle it.

Rain density can change separately from sky darkness: you need to send both packet separately.

src/main/java/net/glowstone/entity/GlowLightningStrike.java
jimmikaelkael: That's it: with a tick living time of 30, lightning flash is sometimes too long (without blinking). When comparing with vanilla thunder effect it's clear that it doesn't have this kind of very long flashes.

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -401,7 +403,7 @@ public void pulse() {
             GlowEntity entity = it.next();
             boolean withinDistance = !entity.isDead() && isWithinDistance(entity);
 
-            if (withinDistance) {
+            if (withinDistance || entity instanceof GlowLightningStrike) {
```
jimmikaelkael, line 406: That was my problem, the lightning entity was removed, then re-added below and so on: the lightning was spawned 29 times.

src/main/java/net/glowstone/entity/GlowLightningStrike.java
jimmikaelkael: @SpaceManiac @turt2live You were perfectly right it's all handled on the client: I was fooled by a bug (look at my latest comment) that did added and removed the lightning entity on every tick.

SpaceManiac:
Merged in f8608d8202d8aee26ea4ec93ce211f0255dd1da2.

Issue #491, opened 2014-10-26
Whitelist enabled state not persisted


Cadiducho:
If you enable the whitelist by command; all is Ok and work successfully but... when you restart the server; the whitelist is disabled by default. 

In addition, when you use '/whitelist on', the file Glowstone.yml remains unchanged (Boolean server.whitelisted)

jengel3:
> In addition, when you use '/whitelist on', the file Glowstone.yml remains unchanged (Boolean server.whitelisted) 

I can't remember, but isn't this default Minecraft behavior?

turt2live:
Vanilla behaviour is to enable (or disable) the whitelist through the server.properties file, therefore enabling (or disabling) the whitelist for the next startup.

Cadiducho:
Tested in Vanilla and Spigot (1.7.10 both), when you execute that command in his respective configuration file the Server change the property, and works fine on restart because of it reads the whitelist status from file, and obviously not from a non static java variable.
Glowstone dont change this property. Perhaps 'you' have to override thr command from Bukkit API

bendem:
Caused by [Server#setWhitelist(boolean)](https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/GlowServer.java#L1299) simply setting a boolean in Glowstone while Vanilla/CB updates the server.properties file as well in that method.

SpaceManiac:
Fixed in a128bc245db36478cf496dba6ea3aa52de04b91c.

PR #501, opened 2014-11-02
Generate default glowstone.yml from ServerConfig.Key

bendem:
_No description provided_

---

_Edited by turt2live_

**Related Links:**
- None known. PR created due to support requests in `#glowstone` relating to configuration options.
- No ticket (none needed)


SpaceManiac:
Do you think it would be reasonable to add new keys added to ServerConfig after the creation of the default `glowstone.yml` to the configuration? This makes things more complicated in that the configuration is now sometimes modified by the server, but this might be useful anyways for things like #491.

src/main/java/net/glowstone/util/ServerConfig.java
```diff
@@ -165,7 +169,7 @@ public void load() {
         }
 
         // if we just created defaults, attempt to migrate
-        if (!exists && migrate()) {
+        if (migrate()) {
```
dequis: Pretty sure this won't run at all if we just created defaults, since there's a `return` above, right below `createDefaultConfig`. You might want to put the migration stuff in that function

bendem:
@dequis that should work better
@SpaceManiac would mean checking for missing config keys when loading the file. That's totally possible, do you want me to add that? Also, do you want me to add `set(Key, Object)` and `save()` to ServerConfig?

bendem:
@m3rcuriel added set and save capabilities to `ServerConfig` in #505.
I'll take a look at adding not found keys in the config if you want me to @SpaceManiac

m3rcuriel:
@bendem Do you want me to edit my PR to have a separate save function rather than relying on the extended class?

bendem:
I added it anyway, can be removed if not needed.

src/main/java/net/glowstone/util/ServerConfig.java
```diff
@@ -283,6 +302,33 @@ private boolean migrate() {
         SHUTDOWN_MESSAGE("server.shutdown-message", "Server shutting down.", Migrate.BUKKIT, "settings.shutdown-message"),
         USE_JLINE("server.use-jline", true),
 
+        // game props
```
turt2live, line 325: Any particular reason to move this entire block of constants?

src/main/java/net/glowstone/util/ServerConfig.java
```diff
@@ -143,38 +147,53 @@ public void load() {
...
-        try {
-            config.load(configFile);
-        } catch (IOException e) {
-            GlowServer.logger.log(Level.SEVERE, "Failed to read config: " + configFile, e);
-        } catch (InvalidConfigurationException e) {
-            report(configFile, e);
+            // attempt to migrate
+            if (migrate()) {
+                GlowServer.logger.info("Migrated configuration from previous installation");
+            }
```
turt2live, line 194: Should we warn for failed migration? Or am I an idiot because that is displayed elsewhere...?

src/main/java/net/glowstone/util/ServerConfig.java
bendem, line 325: I thought having most used config keys first and advanced config last made more sense.

src/main/java/net/glowstone/util/ServerConfig.java
bendem, line 194: It is reported line 201-203 of the new diff

bendem:
Rebased on top of master (only conflict was the addition of `Key.RESOURCE_PACK_HASH`)

src/main/java/net/glowstone/util/ServerConfig.java
turt2live, line 325: I don't think it's a change that needs to be in the PR

bendem, line 325: Reverted

bendem, line 325: Would this change alone fit another PR tho?

I included it because it was related to the automatic generation of the config addressed in this PR and wouldn't break anything since none of the glowstone codebase relies on the ordering of the `ServerConfig.Key` enum.

turt2live, line 325: wait, this affects the generation of the config? If so, then it should be the way it was before. I thought you were mentioning that you preferred it this way.

bendem, line 325: The new config file is ordered based on the enum natural ordering. (my mistake, I thought "Reorder ServerConfig.Key to hide advanced config at the end of the file" was clear enough)

dequis:
Revert "Revert "Revert "what are you even doing"""

bendem:
https://github.com/GlowstoneMC/Glowstone/pull/501#discussion_r20408776

turt2live:
Green-lighting this for @SpaceManiac's review. The order is somewhat questionable to me, but the code functions as intended.

turt2live:
@SpaceManiac This pull request is still pending pull. The pull request has been re-tested and is confirmed to be operational.

PR #503, opened 2014-11-03
Fix placement issues with red sandstone stairs / slabs

dequis:
For some reason mojang decided to use new item IDs for these - they just needed to be added to ItemTable, and include them in a few conditions in BlockSlab.

To test:
- Stairs: just place them in any direction or inverted. Old behavior sticks them with a single orientation.
- Slabs: place two red sandstone slabs on top of each other, they should stack into a double slab. Old behavior leaves two single slabs, the second one floating above with a space in the middle.

---

_Edited by turt2live_

**Related Links:**
- Ticket: #285


turt2live:
Closing as this will be better resolved in #467 (by accident). Thanks for your contribution!

PR #505, opened 2014-11-04
Add config setters and setWhitelist

### LINKED ISSUE #491: Whitelist enabled state not persisted

m3rcuriel:
This PR fixes #491. It adds setters to ServerConfig, and uses those setters in GlowServer to set whitelist in Glowstone.yml.

---

_Edited by turt2live_

**Related Links:**
- Ticket: #491

### DIFF
--- a/src/main/java/net/glowstone/util/ServerConfig.java
+++ b/src/main/java/net/glowstone/util/ServerConfig.java
@@ -64,6 +64,37 @@ public final class ServerConfig {
     }
 
     ////////////////////////////////////////////////////////////////////////////
+    // Value setters
+
+    private void set(String path, Object value) {
+        config.set(path, value);
+        try {
+            config.save(configFile);
+        } catch (IOException e) {
+            GlowServer.logger.log(Level.SEVERE, "Failed to write config: " + configFile, e);
+            return;
+        }
+    }
+
+    public void setString(Key key, String value) {
+        Validate.notNull(value);
+        parameters.put(key, value);
+        set(key.path, value);
+    }
+
+    public void setInt(Key key, int value) {
+        Validate.notNull(value);
+        parameters.put(key, value);
+        set(key.path, value);
+    }
+
+    public void setBoolean(Key key, boolean value) {
+        Validate.notNull(value);
+        parameters.put(key, value);
+        set(key.path, value);
+    }
+
+    ////////////////////////////////////////////////////////////////////////////
     // Value getters
 
     public String getString(Key key) {
--- a/src/main/java/net/glowstone/util/ServerConfig.java
+++ b/src/main/java/net/glowstone/util/ServerConfig.java
@@ -1,5 +1,6 @@
 package net.glowstone.util;
 
+import com.sun.org.apache.xpath.internal.operations.Bool;
 import net.glowstone.GlowServer;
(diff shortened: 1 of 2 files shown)


turt2live:
@m3rcuriel Please resolve the permission changes on the files.

src/main/java/net/glowstone/util/ServerConfig.java
```diff
@@ -64,6 +64,34 @@ public ServerConfig(File configDir, File configFile, Map<Key, Object> parameters
...
+            return;
+        }
+    }
+
+    public void setString(Key key, String value) {
+        Validate.notNull(value);
+        set(key.path, value);
+    }
+
+    public void setInt(Key key, int value) {
```
turt2live: `setInt` and `setString` are not used by this PR - any reason why they were added?

src/main/java/net/glowstone/util/ServerConfig.java
```diff
@@ -64,6 +64,34 @@ public ServerConfig(File configDir, File configFile, Map<Key, Object> parameters
     }
 
     ////////////////////////////////////////////////////////////////////////////
+    // Value setters
+
+    private void set(String path, Object value) {
+        config.set(path, value);
+        try {
+            config.save(configFile);
```
turt2live, line 70: Do we really need to save every change? Currently we don't have anything that does this, but here is a case where this would be undesirable:

```
for(int i=0; i<10000; i++){
    theConfig.set("something."+i, i); 
}
```

m3rcuriel:
I added setInt() and setString() as parallels for the already existing getters. They are unnecessary for the fix that this PR offers, but offer functionality to be used by other PRs, again in parallel to the existing getters. Easy to remove, though.

Regarding saving every change, I now see that there are places where this is undesirable. My initial thought was that saving in the special case of setting whitelist is a good idea in case of crash, so I think I'll put save in a seperate method and call it from GlowServer.

turt2live:
That sounds a bit safer. As for the additional methods, I think they can stay there. They don't harm anything.

dequis:
> Regarding saving every change, I now see that there are places where this is undesirable

Do these exist? Because IMO it's better to avoid requiring a `save()` call after every `set()`, if the config stays at the scale that it is currently (small, no adding entries through iteration, etc)

turt2live:
There are no known severe cases, but it's preventive measures in the event it occurs later. And to be honest it's kind of a workaround and not really valid. I would prefer the additional save call in GlowServer.

src/main/java/net/glowstone/util/ServerConfig.java
```diff
@@ -1,5 +1,6 @@
 package net.glowstone.util;
 
+import com.sun.org.apache.xpath.internal.operations.Bool;
```
dequis: u wot

(this is making checkstyle fail with unused import)

SpaceManiac:
Merged in a128bc245db36478cf496dba6ea3aa52de04b91c.

Issue #510, opened 2014-11-05
java.io.IOException: Broken Pipe from client on whitelist disconnect


m3rcuriel:
When trying to connect to a server that has whitelist enabled, as a player not on the whitelist, you recieve a Broken Pipe error from the client. Possibly caused by Glowstone either a) not sending a graceful disconnect packet, or b) Glowstone sending the packet at the wrong time.

Steps to reproduce:
1) Start server
2) Set whitelist on
3) Connect to server

![00000000-000000](https://cloud.githubusercontent.com/assets/9262491/4911745/e85db2ac-6492-11e4-9312-248b195d1bd4.png)

turt2live:
A client log may be helpful

m3rcuriel:
Client log shows nothing unusual last time I checked. Let me go get that.

turt2live:
Cannot reproduce at the moment:
![img](http://i.turt2live.com/dVdo.png)

m3rcuriel:
Okay. I've tested this more. The issue still exists (I believe) within a larger issue. Across two of my own computers, running two different compiled jars and the downloaded jar, as well as any testing done by gratimax, I've recieved a multitude of sporatic errors, ranging from Broken Pipe, to connection reset by peer, to the expected "you are not whitelisted."

jimmikaelkael:
Please see #373, I've closed the PR because of lack for time to fix this but I think it's all related to the same problem:
The JoinGame packet is sent in GlowPlayer constructor which should deferred to after we have ensured that the player can effectively join.

@m3rcuriel it seems like on my PR that turt2live never had this problem, are you like me using Linux ?

m3rcuriel:
@jimmikaelkael: I've tested client and server, compiled and downloaded, LAN and localhost, on two different networks with radically different latency, on Windows 8, OS X Mavericks, and Arch Linux. Sporadic errors that are apparently random. Feel free to lump this under a larger issue, though. @gratimax and I are trying to diagnose the various connection issues, so this issue is a bit unnecessary.

PR #518, opened 2014-11-07
Fixes player joining game before being kicked for whitelist/ban/full

m3rcuriel:
This PR fixes issue #510 , and #156, as well as touches on issues addressed in PR #373 about errors when not on a server whitelist.

The PR moves the method calls that send the join packet to the player from the constructor to a seperate method, and then calls that method once the player has been confirmed, and the PlayerLoginEvent has not been disallowed.

The creation of the InventoryMonitor is necessary inorder to prevent errors with pulse.

Sorry for putting this on master, I've been having problems with git all day and this is my third time writing this code, and this time I forgot to switch branches.


m3rcuriel:
Closing this PR so that I can get it on the right branch. Sorry for any sort of spam that happens because of this.

PR #521, opened 2014-11-07
Fix the player 'joining' when they should be kicked

m3rcuriel:
ORIGINALLY PR #518

This PR fixes issue #510, and #156, as well as touches on issues addressed in PR #373 about errors when not on a server whitelist.

The PR moves the method calls that send the join packet to the player from the constructor to a seperate method, and then calls that method once the player has been confirmed, and the PlayerLoginEvent has not been disallowed.

The creation of the InventoryMonitor is necessary in order to prevent errors with pulse.

---

_Edited by turt2live_

**Related Links:**
- Original PR: #518 
- Ticket: #510 
- Ticket: #156 
- Ticket: #296 
- Includes changes from #373

### DIFF
--- a/src/main/java/net/glowstone/net/GlowSession.java
+++ b/src/main/java/net/glowstone/net/GlowSession.java
@@ -69,6 +69,10 @@ public final class GlowSession extends BasicSession {
     private InetSocketAddress address;
 
     /**
+     * The state of the connection
+     */
+    private boolean online = false;
+    /**
      * The verify token used in authentication
      */
     private byte[] verifyToken;
@@ -243,6 +247,14 @@ public final class GlowSession extends BasicSession {
     // Player and state management
 
     /**
+     * Get session online state
+     * @return true if this session's state is online
+     */
+    public boolean isOnline() {
+        return online;
+    }
+
+    /**
      * Gets the player associated with this session.
      * @return The player, or {@code null} if no player is associated with it.
      */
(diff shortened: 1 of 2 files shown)


turt2live:
This still displays a message that the player left the server. Does not fix #156 as claimed :(

dequis:
Does this attempt to fix #296?

turt2live:
@dequis yes, as per the description it includes similar changes to #373

m3rcuriel:
No, it is not intended to fix #296. #373 included changes that fixed issue #156, which are included in this PR.

turt2live:
Your PR description says it includes changes from #373 which is a direct fix for #296...

src/main/java/net/glowstone/net/GlowSession.java
```diff
@@ -456,6 +473,8 @@ private void updatePipeline(String key, ChannelHandler handler) {
         getChannel().pipeline().replace(key, key, handler);
     }
 
+
```
turt2live: Extraneous whitespace (x2)

src/main/java/net/glowstone/net/GlowSession.java
```diff
@@ -479,8 +498,10 @@ public void onDisconnect() {
         GlowServer.logger.info(player.getName() + " [" + address + "] lost connection");
 
         final String text = EventFactory.onPlayerQuit(player).getQuitMessage();
-        if (text != null && !text.isEmpty()) {
-            server.broadcastMessage(text);
+        if (online) {
+            if (text != null && !text.isEmpty()) {
```
turt2live: Why not change the if statement to be `online && text != null && !text.isEmpty()` ?

turt2live:
@SpaceManiac This pull request is still pending pull. The pull request has been re-tested and is confirmed to be operational.

PR #529, opened 2014-11-09
Implement Titles API

turt2live:
Through implementing Glowkit#51, this provides plugins (and players) the ability to set titles to players. These titles are fancy things that annoy me personally.

**What does this PR include?**
- Titles (with color code support!)
- `/title`

**What does this PR not include?**
- A proper chat component system
  - A workaround system has been implemented because of how the protocol works. The client is expecting a JSON string and it should be correctly formatted. Shoving messages with spaces (ie: `Hello there!`) will break the client. 
  - Extra checks were done in the command handler to ensure that the player cannot break the client with invalid JSON as well. This performs very basic checks as the client is somewhat resilient to extra keys.

I strongly suggest that when the chat component system is implemented that this gets fixed alongside it. 

---

_Edited by turt2live_

**Related Links:**
- Glowkit: GlowstoneMC/Glowkit#51
- [Sample plugin](https://drive.google.com/folderview?id=0B6JIYSBkLkAnMC13ZVZqekpuRjA&usp=sharing&tid=0B6JIYSBkLkAnTTZYd3Z4ZTMzV3c)


src/main/java/net/glowstone/command/TitleCommand.java
```diff
@@ -0,0 +1,149 @@
...
+                            }
+
+                            JSONArray extra = (JSONArray) extraRaw;
+                            if (extra.size() == 0) {
+                                sender.sendMessage(ChatColor.RED + "Invalid JSON: Could not parse, invalid format?");
+                                return false;
+                            }
+                            for (Object component : extra) {
+                                if (!(component instanceof JSONObject) || !((JSONObject) component).containsKey("text")) {
+                                    sender.sendMessage(ChatColor.RED + "Invalid JSON: Could not parse, invalid format?");
```
dequis: This line is copypasted 5 times and it makes me sad :(

Just an idea: create an exception just for these? Wish it could be more ad-hoc than defining a whole class.

src/main/java/net/glowstone/command/TitleCommand.java
```diff
@@ -0,0 +1,149 @@
...
+                StringBuilder message = new StringBuilder();
+
+                for (int i = 2; i < args.length; i++) {
+                    if (i > 2) message.append(" ");
+                    message.append(args[i]);
+                }
+
+                String raw = message.toString().trim();
+
+                if (raw.contains(" ")) {
```
dequis: This would fail with `{"text":"derp"}` or even `"derp"` and `'derp'`, all of these get displayed as `derp` and have no spaces

src/main/java/net/glowstone/command/TitleCommand.java
turt2live: The parsing is a **temporary replacement** pending a proper system.

src/main/java/net/glowstone/command/TitleCommand.java
turt2live: I don't think the parsing needs to be complicated for all cases right now. It's a **temporary replacement** pending a better system.

src/main/java/net/glowstone/net/message/play/game/TitleMessage.java
```diff
@@ -37,6 +46,87 @@ public TitleMessage(Action action) {
...
+            return "";
+        }
+
+        // TODO: Replace this with proper chat formatting components at some point
+        Object parsed = JSONValue.parse(message);
+        if (parsed == null) {
+            Map<String, Object> converted = null;
+
+            List<Object> extra = new ArrayList<>();
+            String[] parts = message.split(ChatColor.COLOR_CHAR + "");
```
dequis: Aren't you duplicating `glowstone.util.TextMessage.convert` here?

src/main/java/net/glowstone/command/TitleCommand.java
```diff
@@ -0,0 +1,233 @@
...
+            sender.sendMessage(ChatColor.RED + "Usage: " + usageMessage);
+            return false;
+        }
+
+        Player player = Bukkit.getPlayerExact(args[0]);
+
+        if (player == null || (sender instanceof Player && !((Player) sender).canSee(player))) {
+            sender.sendMessage("There's no player by that name online.");
+            return false;
+        } else {
```
dequis: This else isn't needed now, since it's after a branch that returns.

dequis:
FWIW, vanilla shows part of the exception message to the client when parsing json fails. I don't think it's worth bothering, though. Up to you.

turt2live:
There are no exceptions, so I don't think we can show the message :P

src/main/java/net/glowstone/command/TitleCommand.java
```diff
@@ -0,0 +1,233 @@
...
+    private boolean validJson(String raw) {
+        Object object = JSONValue.parse(raw);
+
+        if (object == null) {
+            // Could not parse JSON: Check to see if it's at least a single word
+
+            // Rule set:
+            // 1. Cannot contain a space (or else the client fails)
+            // 2. Must not look like JSON (first character check is sufficient)
+            return !raw.contains(" ") && raw.charAt(0) != '{' && raw.charAt(0) != '[';
```
dequis, line 135: Kinda nitpicky, but `"derp"` and `'derp'` are valid json too, and get displayed as `derp` without the quotes.

turt2live, line 135: You mean `/title dequis title "derp"` ?

src/main/java/net/glowstone/command/TitleCommand.java
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        if (!(object instanceof JSONObject))
+            return false; // Not a valid JSON object
+
+        JSONObject json = (JSONObject) object;
+
+        // Needs a text component at least
+        if (!json.containsKey("text") || !(json.get("text") instanceof String))
```
dequis, line 144: Is there a need to validate this? (to avoid client crashes, for example?)

Because, for example, `["1", "2", {"text": "3"}]` displays as "123" and is perfectlyvalid, but would get rejected by this. There's also `{"selector": "@a"}` (replaced by players) `"translate"`, `"score"`, see here for more details:

http://wiki.vg/Chat#Current_system_.28JSON_Chat.29

So IMO it's better to not validate this.

src/main/java/net/glowstone/command/TitleCommand.java
dequis, line 135: Yep

src/main/java/net/glowstone/command/TitleCommand.java
turt2live, line 144: I do not feel as though all the small cases need to be handled at this time. There is a very clear note that the current implementation is pending a replacement by a better system (chat components). The implementation provided is for demonstration purposes and is really only provided to assist with the testing of the API and protocol functionality.

src/main/java/net/glowstone/command/TitleCommand.java
turt2live, line 135: https://github.com/GlowstoneMC/Glowstone/pull/529/files#r20199528

src/main/java/net/glowstone/command/TitleCommand.java
dequis, line 144: Soooo... better remove the validation? Since a client crash would be a small, rare case here.

turt2live, line 144: The validation is present to avoid errors on the client-side. It's not very good but it saves users from reporting issues relating to titles. It's a bit stricter than vanilla, but it's better than leaving a wide open bug in the PR.

dequis, line 144: Eh, sure, but are you saying you actually found errors on the client side while testing this? Because I was under the assumption that json chat was particularly tolerant to bullshit, since it deals with untrusted input fairly often. But if you got to break it, maybe it's not the client that is protected against the untrusted input.

turt2live, line 144: Before all of this validation I thought the same thing and when testing I simply thought to send "hello there". It kills the client off with a disconnect due to an exception because `hello there` was not valid JSON. The vanilla server does a much better job than what is done here, but that is because it has actual classes to serialize into and therefore validate. 

**TL;DR** The client hates you if you send it bullshit.

src/main/java/net/glowstone/command/TitleCommand.java
```diff
@@ -0,0 +1,233 @@
...
+    /**
+     * Converts a valid JSON chat component to a basic colored string. This does not parse
+     * components like hover or click events. This returns null on parse failure.
+     *
+     * @param json the json chat component
+     *
+     * @return the colored string, or null
+     */
+    // TODO: Replace with proper chat components when possible
+    public static String convertJson(JSONObject json) {
```
SpaceManiac, line 180: Judging by the docs, this can be replaced with `new TextMessage(json).flatten()`, which additionally supports non-color formatting.

src/main/java/net/glowstone/net/message/play/game/TitleMessage.java
```diff
@@ -37,6 +41,26 @@ public TitleMessage(Action action) {
...
+
+    public static TitleMessage fromOptions(TitleOptions options) {
+        return new TitleMessage(Action.TIMES, options.getFadeInTime(), options.getVisibleTime(), options.getFadeOutTime());
+    }
+
+    private static String toValidString(String s) {
+        if (s == null) return "";
+
+        JSONObject parsed = TextMessage.convert(s);
+        if (parsed == null) return "";
```
SpaceManiac: `convert()` never returns null. `TextMessage.convert(s).toJSONString()` can be invoked through `new TextMessage(s).encode()` without having to additionally expose `convert`.

It might also make sense to make `text` a `TextMesssage` instead, which is the class acting as the placeholder for whatever eventually becomes the chat model.

src/main/java/net/glowstone/command/TitleCommand.java
dequis, line 180: @turt2live seems like you forgot to address this one

turt2live:

turt2live:
Closing this PR - Project has moved on to a [new repo](https://github.com/GlowstoneMC/Glowstone). 

If this needs implementing in the newer repository I'll try and find time to implement it.

PR #531, opened 2014-11-09
Add sponge block

socram8888:
This PR adds a general-purpose taxicab block iterator and employs it to add the block behaviour of sponges in 1.8.

Sponge working

### DIFF
--- a/src/main/java/net/glowstone/block/ItemTable.java
+++ b/src/main/java/net/glowstone/block/ItemTable.java
@@ -133,6 +133,7 @@ public final class ItemTable {
         reg(Material.STATIONARY_LAVA, new BlockLava());
         reg(Material.STANDING_BANNER, new BlockBanner());
         reg(Material.WALL_BANNER, new BlockBanner());
+        reg(Material.SPONGE, new BlockSponge());
 
         reg(Material.SIGN, new ItemSign());
         reg(Material.REDSTONE, new ItemPlaceAs(Material.REDSTONE_WIRE));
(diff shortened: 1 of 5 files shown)


socram8888:
How did I manage to compile an Iterator without the remove method is beyond my own comprehension.

Tonodus:
@socram8888 it's by default empty. (Java 8 default methods for interfaces)

turt2live:
Incorrect drops for wet sponge. Also the naming of the iterator is kinda funny and doesn't seem to really apply to the definition of a taxicab number. I'm not sure on what else to call it though

socram8888:
I called it TaxicabBlockIterator because in [MinecraftWiki](http://minecraft.gamepedia.com/Sponge), it says that "A sponge absorbs water around itself (water source blocks or flowing water) out to a **taxicab** distance of 7 in all directions". I'm open to suggestions, though.

socram8888:
Drops have been fixed.

dequis:
This fixes [#498](https://github.com/SpaceManiac/Glowstone/issues/498?)

<sub><sub>sorry i couldn't resist</sub></sub>

turt2live:
This PR fails on edge cases. It also doesn't soak up the correct number of blocks.

Edge case:
edge

socram8888:
That's how it's supposed to work, it will only scan if at least one **face** has water. That's why the wiki also says "taxicab", because it won't work for diagonals.

socram8888:
This last PR not only improves performance but also it now absorbs the correct amount of water blocks.

SpaceManiac:
Merged in 6e86a71dd994a867e3bbb34db09f8d8af2d8feb6.

PR #539, opened 2014-11-12
Drop attached blocks when the block it's attached to is destroyed by a player.

jimmikaelkael:
OK, I know the code is currently awful. I'd like to have some point of views as I'm not sure the proper way to handle this would be in BlockType.


dequis:
Isn't this essentially the same as BlockNeedsAttached from #249?

jimmikaelkael:
The same ? No judging by the length #249 is far more complete :)
Otherwise you're true @dequis, that's it.

Closing this, sorry for the trouble.

ZephireNZ:
Good lord that is a copy-paste mess. You'd be better off creating an array/list of the directions, and then iterating over that.

jimmikaelkael:
Did you read my 1st comment ? No it seems...
Anyway this is closed in favor of Tonodus's Block Physics PR.

PR #540, opened 2014-11-12
Drop the correct drops

### LINKED ISSUE #527: Bad drop mappings
This is an extension of #98 so that it doesn't get cluttered.

Known/confirmed cases. Checked means 'resolved in master'.
- _Coming Soon_

_Update: List is being generated [here](https://docs.google.com/spreadsheets/d/1Qgky1a2zsUeKfOblpNLmmVF95QmkKd_bh0tvuu143tA/edit#gid=0)_

Tonodus:
This should fix #527 expect that blocks with hardness 0 won't be dropped (redstone, redstone torches, torches, flowers, etc).

I'm not 100% sure I've added all known cases, I will go through any cases when #527 or [this](https://docs.google.com/spreadsheets/d/1Qgky1a2zsUeKfOblpNLmmVF95QmkKd_bh0tvuu143tA/edit?pli=1#gid=0) is updated.

### DIFF
--- a/src/main/java/net/glowstone/block/ItemTable.java
+++ b/src/main/java/net/glowstone/block/ItemTable.java
@@ -100,6 +100,7 @@ public final class ItemTable {
         reg(Material.LAPIS_BLOCK, new BlockDirectDrops(ToolType.STONE_PICKAXE));
         reg(Material.QUARTZ_ORE, new BlockDirectDrops(Material.QUARTZ, ToolType.PICKAXE));
         reg(Material.REDSTONE_ORE, new BlockRandomDrops(Material.REDSTONE, 0, 3, 4, ToolType.IRON_PICKAXE));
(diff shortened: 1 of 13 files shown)


turt2live:
`PRHOLD: Mapping of vanilla values. See doc`

turt2live:
@Tonodus You broke the tests with your last commit.

Tonodus:
@turt2live dammit... fixed.

turt2live:
@Tonodus This pull request is no longer mergeable. Please resolve the issue as soon as possible, thanks!

dequis:
The hold for this one isn't really needed, is it? Because I suspect it might be a while until the spreadsheet covers absolutely everything

turt2live:
I actually completely forgot about that spreadsheet. I'll work on a better way to generate a list of values though because manual methods are... hard.

PR #541, opened 2014-11-12
Material values: hardness, blast-resistance, etc.

### LINKED ISSUE #384: Blocks with hardness set to 0 never really break, and doesn't drop.

Tonodus:
This PR fixes #384 and is an alternative for #385.

**What's WIP:**
The code is 100% finished and ready for review, only the concrete values must be filled in. I'll start on doing this, when it's said that the code is correct.

For testing: torches drop items now, since they have hardness == 0 and insta-break therefore.

This PR should be easily extendable for other values as well (f.e. light-level). It could also replace hardcoded values if wanted (f.e. whether blocks can absord). If there's the need, a Glowkit-PR could be created, allowing plugins to fetch these values as well.

---

Edit by turt2live: This, due to it's inheritance of #385, also references #372


turt2live:
Please don't modify the TNT PR, it's just safer to not do that.

Tonodus:
No problem, won't do that then.

dequis:
Fuck yeah yaml.

turt2live:
@Tonodus What is the purpose for the large changeset? It doesn't look expandable like you claim and it seems like it's overcomplicating the fix to be honest. A description detailing what is going on in this PR would also be appreciated as the code is not super clear from an internal perspective (what it does is obvious, but the BuiltinMaterialManager class is somewhat overcomplicated and unclear as to what it is supposed to do. It looks like it does 12 things when it only needs 1).

Tonodus:
@turt2live Yeah, it's actually a bit more complicated than I planned it to be.
However, the idea was to have an immutable Value class (`MaterialValueManager$Value`).
To build an object of this immutable class, there need to be some sort of Builder (`BuiltinMaterialValueManager$ValueBuilder`). That takes care of setting default values for the Value-Object, and change the default values to the concrete ones, if the're found in the `materialValues.yml` file.
To easily handle that mentioned file, I added `BuiltinMaterialValueManager$ValueKey`. That enum searches a given ConfigurationSection to update a given `ValueBuilder`.

So what does make this such complicated:
- Having final values in the Value class
- Having the values fixed in the class: `float hardness;` instead of some dynamic fetching from a ConfigurationSection
- Not using sort of Keys: alternative `get(ValueKey.HARDNESS)`

Why did I do so? Basically the idea it to use immutable objects and not fetching things dynamically like ServerConfig.java does, f.e. The benefits would be best performance for fetching values and concrete return-values. (What would getInt(KEY.HARDNESS) return?).

**Edit:** Pushed a version storing the ConfigurationSection behind it's values. The BuiltinMaterialValueManager becomes much smaller then.

turt2live:
So you've re-implemented yaml configurations. If I remember correctly, the YAML library Glowstone uses actually does memory caching so that the value retrieved from `get____` is from memory, and not disk. I still don't see the benefits of such a large class to handle something as simple as "get this value".

Tonodus:
@turt2live Have you seen the latest diff? It's not that large and IMO small enought.

> So you've re-implemented yaml configurations.

I don't think so. I wrote a wrapper about a specific yaml-configuration.

turt2live:
I'm not so sure the approach is ideal, but it looks better. I may poke it a bit and PR something to you if it works out.

src/main/java/net/glowstone/block/MaterialValueManager.java
```diff
@@ -0,0 +1,26 @@
...
+    Value getValue(Material material);
+
+    public interface Value {
+        /**
+         * Returns the hardness-component of this value.
+         * @return returns the hardness (or Float.MAX_VALUE for infinity hardness)
+         */
+        float getHardness();
+
+        float getBlastResistance();
```
turt2live, line 28: This should also have documentation on it

src/main/java/net/glowstone/block/GlowBlock.java
```diff
@@ -230,6 +232,10 @@ public boolean isLiquid() {
         return mat == Material.WATER || mat == Material.STATIONARY_WATER || mat == Material.LAVA || mat == Material.STATIONARY_LAVA;
     }
 
+    public MaterialValueManager.Value getMaterialValues() {
```
turt2live: Pluralized methods should return collections or be renamed as per their return type.

src/main/resources/builtin/materialValues.yml
```diff
@@ -0,0 +1,207 @@
+default:
+  hardness: 1
+  blastResistance: 0
+values:
+  AIR:
```
turt2live: The manager should be written in a way where you should not have to declare empty sections for every block that uses a default value, please correct this.

Tonodus: It is already. I just list all blocks, so that no value is forgotten. When all values are inserted, blocks without non-defaults can be deleted.

turt2live: Why can't the values be inserted now and cleaned up now?

dequis: From the description:

> The code is 100% finished and ready for review, only the concrete values must be filled in. I'll start on doing this, when it's said that the code is correct

turt2live: Part of ensuring the code is working is to test the differing cases of blocks. I'd also not like to extend the review process by re-testing this pull request over and over for table updates.

turt2live:
@Tonodus Your pull request is now unmergeable due to recent commits on the master branch. Please resolve the conflicts as soon as possible. Please also address the concerns mentioned in this pull request previously.

src/main/java/net/glowstone/block/GlowBlock.java
Tonodus: I renamed `Value` to `ValueCollection` instead. Imo this makes more sense, since it's not a single value, bot the material's collection of values.

src/main/resources/builtin/materialValues.yml
Tonodus: I added a few more values for blastResistance now, so there shouldn't any problem testing this. I'm convinced, there's no need to test every block's blast resistance.
I also inserted all `hardness: 0` values for testing, since other values do not affect glowstone in any way (now).
The file will be completed asap.

Issue #486, opened 2014-10-25
Teleporting players over worlds creates artifacts


turt2live:
Using [DumbTeleport](http://dev.bukkit.org/bukkit-plugins/dumbteleport/) (a simple teleport plugin) that uses the Bukkit API to teleport players ([source](https://github.com/turt2live/DumbTeleport/blob/master/src/main/java/com/turt2live/dumbtp/DumbTp.java#L199)), you can create artifacts of players if you teleport them over worlds. (eg: world -> nether -> world again)

Image of failure:
![image](https://cloud.githubusercontent.com/assets/1190097/4780561/5ab54e78-5c6e-11e4-903b-98327cacb3b1.png)

Issue #447, opened 2014-10-17
Player visually hangs around after death


earthwolfgirl:
I have build 91 installed on my server and My mod informed me there was two of me. I have attached images
![duplication](https://cloud.githubusercontent.com/assets/9288561/4686318/28441bca-5644-11e4-87b3-957bbf61ef80.jpg) 
![afterimage](https://cloud.githubusercontent.com/assets/9288561/4686319/2cc22a2a-5644-11e4-8efd-e84245385729.jpg)

turt2live:
Reproduction steps? What happened to cause this? What happened prior to this?

dequis:
I'm betting it's related to death. The vanilla client would handle something like this by not showing the second one.

What mod is this?

turt2live:
Glowstone doesn't have full death support, so it is more than likely Glowstone if this was caused by death.

earthwolfgirl:
So Apparently my mods decided to  /kill me while i was afk and that what caused the afterimage or duplication or what you want to call it.

turt2live:
I've updated the issue title to better reflect the problem. Thanks!

dequis:
Oh.. mod as in moderator? Derp.

earthwolfgirl:
Sorry I didn't clarify that dequis. I forget that mod is also for modifications as well as moderators.

m3rcuriel:
Hey, I can't seem to duplicate this. Can you give more information to reproduce the bug?

turt2live:
@m3rcuriel One easy way is to get a friend and kill them.

dequis:
Just to be perfectly clear: we are not encouraging anyone to kill their friends IRL.

PR #543, opened 2014-11-14
Do not reassign entity ids on world changes / respawns

Johni0702:
This PR fixes a bug caused by when teleporting entities to other worlds due to which entities are displayed multiple times to the clients.
This is also the cause for players hanging around after death which is also fixed with this PR.
Entities near the respawn location being invisible is also fixed with this PR.
**Current behavior**
Entity ids are allocated on a per world basis in the `EntityManager` and are reassigned if entities change the world or players respawn. This leads to the old id not being removed and artifacts being created. Furthermore the entity id should not change at all (due to CraftBukkit compatibility).
**New behaviour**
Entity ids are now allocated in the `EntityIdManager` of which only a single instance exists per server. This allows the entity to change its world without being reassigned  to a new id by the `EntityManager` of the new world.
This PR also resets all known entities when a player respawns in the same world. This causes all entities to be resend after the respawn which solves the third issue.

**Related Links:**
- Ticket: #486 
- Ticket: #447

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -579,9 +579,9 @@ public final class GlowPlayer extends GlowHumanEntity implements Player {
     private void spawnAt(Location location) {
         // switch worlds
         GlowWorld oldWorld = world;
-        world.getEntityManager().deallocate(this);
+        world.getEntityManager().unregister(this);
         world = (GlowWorld) location.getWorld();
-        world.getEntityManager().allocate(this);
+        world.getEntityManager().register(this);
 
         // switch chunk set
         // no need to send chunk unload messages - respawn unloads all chunks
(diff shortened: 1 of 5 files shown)


turt2live:
Because we have another artifact problem I will be testing that as well. If the fix is at all related, I will expect a fix for it here :)  (death artifacts)

Johni0702:
The death artifacts do indeed have the same reason. The only difference is that players don't get teleported to another world and therefore we have to check whether their id has changed. Previous commit fixed that.
I edited the description to also include that ticket.

turt2live:
Due to CraftBukkit API compatibility, this is an incorrect fix. The entity ID shouldn't be changing, fixing that would be the correct fix. Please make the required adjustments, thanks!

Johni0702:
@turt2live Changed PR content and updated title and description accordingly.

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -629,6 +629,15 @@ public void respawn() {
         // fire event and perform spawn
         PlayerRespawnEvent event = new PlayerRespawnEvent(this, dest, spawnAtBed);
         EventFactory.callEvent(event);
+        if (event.getRespawnLocation().getWorld() == location.getWorld() && knownEntities.size() > 0) {
```
turt2live: Use `.equals`

SpaceManiac:
Merged in 37f01ea44298df86cf45756e59082df4abb60c8c.

PR #550, opened 2014-11-15
Cacti hurt when hugged

jaredStef:
I added a method called canTakeCactusHurt. In the living entity class it always returns true because obviously animals aren't immune to cacti. In the GlowHumanEnity class I overrode the method so it returns true if the player is in survival mode or adventure mode. In the living entity class there is a method called takeCactusDamage. It gets the block the player is standing on as well as the blocks 1 up on each side. If they are cacti and the player canTakeCactiDamage() it hurts them.  In the livingEntity class I add an else clause so that if the no damage ticks aren't greater than zero they are reset so that the player can take damage again. Aaron helped me get set up and helped with the some of the logic.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowHumanEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowHumanEntity.java
@@ -275,7 +275,7 @@ public abstract class GlowHumanEntity extends GlowLivingEntity implements HumanE
 
     @Override
     protected boolean canTakeDamage(EntityDamageEvent.DamageCause damageCause) {
-        return gameMode == GameMode.SURVIVAL || gameMode == GameMode.ADVENTURE && super.canTakeDamage(damageCause);
+        return (gameMode == GameMode.SURVIVAL || gameMode == GameMode.ADVENTURE) && super.canTakeDamage(damageCause);
     }
 
     ////////////////////////////////////////////////////////////////////////////
--- a/src/main/java/net/glowstone/entity/GlowHumanEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowHumanEntity.java
@@ -12,6 +12,7 @@ import org.bukkit.GameMode;
 import org.bukkit.Location;
 import org.bukkit.Material;
 import org.bukkit.entity.HumanEntity;
+import org.bukkit.event.entity.EntityDamageEvent;
 import org.bukkit.event.inventory.InventoryType;
 import org.bukkit.inventory.EntityEquipment;
 import org.bukkit.inventory.Inventory;
@@ -273,8 +274,8 @@ public abstract class GlowHumanEntity extends GlowLivingEntity implements HumanE
     // Health
 
     @Override
-    protected boolean canTakeDamage() {
-        return gameMode == GameMode.SURVIVAL || gameMode == GameMode.ADVENTURE;
+    protected boolean canTakeDamage(EntityDamageEvent.DamageCause damageCause) {
+        return gameMode == GameMode.SURVIVAL || gameMode == GameMode.ADVENTURE && super.canTakeDamage(damageCause);
     }
 
     ////////////////////////////////////////////////////////////////////////////
--- a/src/main/java/net/glowstone/entity/GlowHumanEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowHumanEntity.java
@@ -381,4 +381,9 @@ public abstract class GlowHumanEntity extends GlowLivingEntity implements HumanE
             ((GlowInventory) inventory).removeViewer(this);
         }
     }
+
+    @Override
+    public boolean canTakeCactusDamage() {
+        return gameMode == GameMode.SURVIVAL || gameMode == GameMode.ADVENTURE;
+    }
 }
--- a/src/main/java/net/glowstone/entity/GlowHumanEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowHumanEntity.java
@@ -273,7 +273,7 @@ public abstract class GlowHumanEntity extends GlowLivingEntity implements HumanE
     // Health
 
     @Override
-    protected boolean canDrown() {
+    protected boolean canTakeDamage() {
         return gameMode == GameMode.SURVIVAL || gameMode == GameMode.ADVENTURE;
     }
 
@@ -380,10 +380,5 @@ public abstract class GlowHumanEntity extends GlowLivingEntity implements HumanE
         if (inventory instanceof GlowInventory) {
             ((GlowInventory) inventory).removeViewer(this);
         }
-    }
-
-    @Override
-    public boolean canTakeCactusDamage() {
-        return gameMode == GameMode.SURVIVAL || gameMode == GameMode.ADVENTURE;
     }
 }
--- a/src/main/java/net/glowstone/entity/GlowLivingEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowLivingEntity.java
@@ -483,16 +483,16 @@ public abstract class GlowLivingEntity extends GlowEntity implements LivingEntit
 
         // Gets the block the are standing on and the location of it
         Material blockBelow = getLocation().getBlock().getType();
-        Location blockLocationX = getLocation().getBlock().getLocation();
-        Location blockLocationNegX = getLocation().getBlock().getLocation();
-        Location blockLocationZ = getLocation().getBlock().getLocation();
-        Location blockLocationNegZ = getLocation().getBlock().getLocation();
+        Location blockLocationX = getLocation();
+        Location blockLocationNegX = getLocation();
+        Location blockLocationZ = getLocation();
+        Location blockLocationNegZ = getLocation();
 
         // sets the location to its respective cord
-        blockLocationX.setX(blockLocationX.getX() + 1);
-        blockLocationNegX.setX(blockLocationNegX.getX() - 1);
-        blockLocationZ.setZ(blockLocationZ.getZ() + 1);
-        blockLocationNegZ.setZ(blockLocationNegZ.getZ() - 1);
+        blockLocationX.add(1, 0, 0);
+        blockLocationNegX.subtract(1, 0, 0);
+        blockLocationZ.add(0, 0, 1);
+        blockLocationNegZ.subtract(0, 0, 1);
 
         //Checks if the living entity is near a cactus or is on top of a cactus and hurts them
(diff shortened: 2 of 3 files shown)


src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -614,4 +647,5 @@ public Entity getLeashHolder() throws IllegalStateException {
     public boolean setLeashHolder(Entity holder) {
         return false;
     }
+
```
Aaron1011: You might want to get rid of this extra whitespace.

jaredStef:
I just removed the whitespace in a recent commit. (the one you can see above)

src/main/java/net/glowstone/entity/GlowHumanEntity.java
```diff
@@ -381,4 +381,9 @@ private void removeViewer(Inventory inventory) {
             ((GlowInventory) inventory).removeViewer(this);
         }
     }
+
+    @Override
+    public boolean canTakeCactusDamage() {
```
dequis: move this method right below canDrown? also since it does exactly the same, maybe both could be abstracted to calls to a function like canTakeDamage? maybe.

dequis:
+1 for hugs.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -470,6 +474,35 @@ public void setLastDamage(double damage) {
...
+    }
+
+    public void takeCactusDamage() {
+
+        // Gets the block the are standing on and the location of it
+        Material blockBelow = getLocation().getBlock().getType();
+        Location blockLocationX = getLocation().getBlock().getLocation();
+        Location blockLocationNegX = getLocation().getBlock().getLocation();
+        Location blockLocationZ = getLocation().getBlock().getLocation();
+        Location blockLocationNegZ = getLocation().getBlock().getLocation();
```
dequis: copypaste detected

dequis: also wait how is `getLocation().getBlock().getLocation()` different from just `getLocation()` anyway

jaredStef: Good question... I don't think that it is now that I think about it. Im going to change it to getLocation and test it. If its the same ill push to my fork.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -470,6 +474,35 @@ public void setLastDamage(double damage) {
...
+
+        // sets the location to its respective cord
+        blockLocationX.setX(blockLocationX.getX() + 1);
```
dequis: `whateverlocation.add(1, 0, 0)`?

src/main/java/net/glowstone/entity/GlowLivingEntity.java
jaredStef: and about the copy paste... I had it at the bottom of the class so I copy pasted it so it would be under the health section.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
jaredStef: I'll get on it.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
dequis: on a second thought it might be more 'rounded' than the location of the player entity (which can be sub-block)

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -470,6 +474,30 @@ public void setLastDamage(double damage) {
...
+
+        // Gets the block the are standing on and the location of it
+        Material blockBelow = getLocation().getBlock().getType();
+        Location blockLocationX = getLocation();
+        Location blockLocationNegX = getLocation();
+        Location blockLocationZ = getLocation();
+        Location blockLocationNegZ = getLocation();
+
+        // sets the location to its respective cord
+        blockLocationX.add(1, 0, 0);
```
turt2live: Instead of adding and subtracting relative positions, use a predefined array/collection of BlockFaces and iterate those.

jaredStef:
You can stand near the corner of cacti and not get hurt because the radius is a square but Im working on that...

dequis:
Some generic method to detect hugging would be nice.

jaredStef:
What would be the purpose of that other than for cacti? Its not hard and seems like a good idea, but I honestly don't see any other use cases.

turt2live:
Futurrrreee

sent from mobile

jaredStef:
I guess I could see people wanting to make custom blocks or mini games or things. 1 commit later...
Hmm thinking of a method name...

jaredStef:
I got suffocation working as well but thats for another PR and branch...

jaredStef:
whoops :( didn't mean to close it (clicked wrong button)

dequis:
I was thinking something along the lines of `isPlayerHuggingMaterial`, separating it from the damage checks. Also, while i'm a fan of hugs, that might not be the best name for it.

edit: this isn't just players - `isHuggingMaterial` or maybe `isTouchingMaterial` or whatever.

ZephireNZ:
I prefer 'touching' to 'hugging'. Hugging is ok, but it seems very 'cutesy' more than anything.

dequis:
Note that hugs are a vanilla feature, [see the footer of this page for more details](https://dinnerbone.com/about/)

jaredStef:
lol It can be called whatever. I thought I'd add some flavor and call it hugs but if the majority doesn't want hugs I guess they won't get any lol

PaulBGD:
+1 for not trying to redo logic in the damage method

Tonodus:
You should move isTouchingMaterial up to GlowEntity instead of LivingEntity.

jaredStef:
Note: The cactus damage won't be as accurate as it would be in vanilla because I can't make it that precise until entity support is finished. In vanilla it looks at the bounding box.

ZephireNZ:
Maybe put a TODO, so we know to go back to it once entities are implemented?

jaredStef:
I did.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -277,10 +280,10 @@ protected Sound getDeathSound() {
     }
 
     /**
-     * Get whether this entity should take drowning damage.
-     * @return whether this entity can drown
+     * Get whether this entity should take damage.
+     * @return whether this entity can take damage.
      */
-    protected boolean canDrown() {
+    protected boolean canTakeDamage() {
```
turt2live: For safety, pass in the desired DamageCause to this method. There are a few that should ALWAYS return true, but won't if kept secret like this. The argument won't be used in this PR, but I'd rather fix this now.

src/main/java/net/glowstone/entity/GlowHumanEntity.java
```diff
@@ -273,7 +273,7 @@ public void setOp(boolean value) {
     // Health
 
     @Override
-    protected boolean canDrown() {
+    protected boolean canTakeDamage() {
         return gameMode == GameMode.SURVIVAL || gameMode == GameMode.ADVENTURE;
```
turt2live: This should call super as well. Again, not directly related to this PR, but it will save some work in the future. 

See https://github.com/GlowstoneMC/Glowstone/pull/550/files#r20484036

jaredStef:
I'll get on it

jaredStef:
Why would I call super because I don't always want to return true I only want to do it if they are in adventure or survival, but I want all living entities such as animals to take damage

turt2live:
Because later there will be other reasons why a player (or any entity for that matter) cannot take damage, including invulnerability. A simple check of `super.canTakeDamage() && (adventure || survival)` should be fine.

src/main/java/net/glowstone/entity/GlowHumanEntity.java
```diff
@@ -273,8 +274,8 @@ public void setOp(boolean value) {
     // Health
 
     @Override
-    protected boolean canDrown() {
-        return gameMode == GameMode.SURVIVAL || gameMode == GameMode.ADVENTURE;
+    protected boolean canTakeDamage(EntityDamageEvent.DamageCause damageCause) {
+        return gameMode == GameMode.SURVIVAL || gameMode == GameMode.ADVENTURE && super.canTakeDamage(damageCause);
```
turt2live: This will evaluate incorrectly

turt2live:
Cacti hugging kill timer is inaccurate. It should damage 1 point every half second or so (1 point for every 10 ticks). I don't know the actual numbers, I'm just guessing based off of a quick test on vanilla.

Aaron1011:
I'm found a couple bugs:
- The player is damaged when standing _anywhere_ on a diagonal block. You might want to remove this until entity hitboxes get added.

None of the detection is going to be perfect without entity hitbox support, but being damaged on a diangonal block currently looks even weirder than on a non-diagonal block.

turt2live:
@Aaron1011 The "weird" bounding box problem is not a bug. Because of the lack of true bounding boxes, it is perfectly expected (and desired) to have the behaviour you've mentioned.

Aaron1011:
I know, I stated that in my comment. But if you want to keep it that way for now, it's fine.

turt2live:
And I'm responding to let the author know that the issue is perfectly valid and does not (and should not) need a fix by any means.

Aaron1011:
Okay, then.

jaredStef:
I'm kinda a perfectionist and I personally hate the fact its not perfect, but I guess it will have to do for now. In the future it shouldn't be too hard to fix because all you have to do is change one method. I added a todo comment also that its not finished. I'm going to submit another PR for suffocation and work on other damages stated in Issue #203

jaredStef:
I'm not being impatient or anything but I was wondering why my PR hasn't been merged yet. I have 2 more (void damage, and suffocation) I would like to submit, but I cant because they depend on this.

turt2live:
The secondary review process is lengthy and the assigned reviewer has been busy for the last few weeks. It is currently on the list of things to do along with many other things.

jaredStef:
ok :+1:

SpaceManiac:
Merged in aaa3e8ab224adbdedc5beac2ded8565d3b5d6fa0. Wanted to get the bounding box stuff working because I felt the incredibly large damage zone wasn't working out.

PR #553, opened 2014-11-17
Add Potion ItemMeta

ZephireNZ:
This adds implementation for PotionMeta. Requires GlowstoneMC/Glowkit#58 for the new ShowParticles NBT metadata.

PotionMeta is used on potions to replace the effect potion, and/or add more that one effect to the potion.

### DIFF
--- a/src/main/java/net/glowstone/inventory/GlowItemFactory.java
+++ b/src/main/java/net/glowstone/inventory/GlowItemFactory.java
@@ -124,6 +124,8 @@ public final class GlowItemFactory implements ItemFactory {
                 return new GlowMetaFirework(meta);
             case FIREWORK_CHARGE:
                 return new GlowMetaFireworkEffect(meta);
+            case POTION:
+                return new GlowMetaPotion(meta);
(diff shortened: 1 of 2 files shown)


dequis:
Referencing #77

turt2live:
@ZephireNZ How does this hold up against old worlds/files? Such as a server "upgrading" from 1.7.10 where ShowParticles doesn't exist. Would plugin data be restored (items in YAML files) correctly as well as player inventories/effects?

ZephireNZ:
Sorry for the late reply. Currently the PR just has defaults for non-required parts, if it doesn't exist in NBT:

```
PotionEffectType type = PotionEffectType.getById(tag.getByte("Id"));
int duration = tag.getInt("Duration");
int amplifier = tag.getByte("Amplifier");
boolean ambient = tag.isByte("Ambient") ? tag.getBool("Ambient") : false;
boolean particles = tag.isByte("ShowParticles") ? tag.getBool("ShowParticles") : true;
```

PR #557, opened 2014-11-18
Parse JSON when loading sign text NBT

### LINKED ISSUE #528: Sign text issues (when loading vanilla maps?)
While testing a PR, turt2live loaded a texture pack test map in which the signs had text like this in glowstone, but not in vanilla:

```
null
"Armor"
null
null
```

or 

```
""
"Armor"
""
""
```

Additionally, lines with long text had words disappear from them:

```
""
"Brown
""
""
```

Was supposed to be `Brown Mushroom`. The red mushroom sign next to it displayed as `"Red Musroom"` (sic), so it's probably related to text length.

dequis:
Fixes #528

Still uses plaintext internally, since we don't have a robust json chat parser yet. TextMessage.asPlaintext() is rather primitive too

Saving also uses plaintext for the sake of simplicity. Vanilla handles that correctly, since it's needed when migrating from 1.7 maps.

To test this:
- Start glowstone, place some signs with text and some empty lines
- Load same world in [vanilla](https://s3.amazonaws.com/Minecraft.Download/versions/1.8/minecraft_server.1.8.jar), save (this turns sign text into json)
- Load same world in glowstone

Old behavior adds quotes around lines with content, and either `null` or `""` for lines without.

New behavior should show no differences.

### DIFF
--- a/src/main/java/net/glowstone/net/GlowBufUtils.java
+++ b/src/main/java/net/glowstone/net/GlowBufUtils.java
@@ -258,7 +258,7 @@ public final class GlowBufUtils {
      * @throws IOException on read failure.
      */
     public static TextMessage readChat(ByteBuf buf) throws IOException {
-        return TextMessage.decode(ByteBufUtils.readUTF8(buf));
+        return TextMessage.decode(ByteBufUtils.readUTF8(buf), "parse error");
     }
 
     /**
--- a/src/main/java/net/glowstone/block/entity/TESign.java
+++ b/src/main/java/net/glowstone/block/entity/TESign.java
@@ -7,7 +7,6 @@ import net.glowstone.entity.GlowPlayer;
 import net.glowstone.util.nbt.CompoundTag;
 import net.glowstone.util.TextMessage;
 import org.bukkit.Material;
-import org.json.simple.parser.ParseException;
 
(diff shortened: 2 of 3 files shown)


src/main/java/net/glowstone/net/GlowBufUtils.java
```diff
@@ -258,7 +258,7 @@ public static void writeUuid(ByteBuf buf, UUID uuid) {
      * @throws IOException on read failure.
      */
     public static TextMessage readChat(ByteBuf buf) throws IOException {
-        return TextMessage.decode(ByteBufUtils.readUTF8(buf));
+        return TextMessage.decode(ByteBufUtils.readUTF8(buf), "parse error");
```
dequis, line 261: Since I know someone's going to complain - yes, this is kinda lame. It keeps the behavior from the old `TextMessage#decode(String)` which returned this dumb fallback string when parsing failed.

Suggestions welcome

turt2live:
@dequis or @SpaceManiac if you have a chance, please confirm if this is now not needed due to https://github.com/GlowstoneMC/Glowstone/commit/d07e99d4a2aa66878a7055cc5d21917b0df3808b

turt2live:
`<dx> this kills the PR`

PR #566, opened 2014-11-23
Improved Block/ItemType system

### LINKED ISSUE #536: Use composition for a improved BlockType system
The currently used [BlockType](https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/block/blocktype/BlockType.java)/[ItemType](https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/block/itemtype/ItemType.java) system is extended by inheritance. That's a **big** design mistake IMO and it should be changed asap to a composition based one.

An examples of problem with the current BlockType system:

Same problem when implementing `TrapDoors` (place logic & openable) or `FenceGate`s (directional placing & openable).
You cannot let `BlockOpenable` extends `BlockDoor`, since that would cause FenceGate problems and you cannot let `BlockDoor` extend `BlockOpenable`, since this would cause `BlockIronDoor` to open with right-click... and since we write in java, `BlockWoodenDoor` cannot extend _both_ `BlockDoor` and `BlockOpenable`.

Next example:

Ladder could be replaced by (redstone)torches etc.

Same as above: **Inheritance causes *Diamond of Deaths***

So what about solutions? Instead of _inheritance_ Glowstone should use _composition_!
How _exactly_ that can be achieved should be considerated and asap decided, because almost every added BlockXXX.java needs probably changes when introducing that.

My thoughts:
- Simple composition:

``` java
        class Openable { void onRightClick(GlowBlock b) {...} }

        class Door { void onPlace(GlowBlock b) {...} void onBreak(GlowBlock b) {...} }

        class WoodenDoor extends BlockType {
            private Openable openSystem;
            private Door doorSystem;

            @Override void onRightClick(GlowBlock b) { openSystem.onRightClick(b); }
            @Override void onBreak(GlowBlock b) { doorSystem.onBreak(b); }
            @Override void onPlace(GlowBlock b) { doorSystem.onPlace(b); }
        }
```

**Pros:** easy to implement
**Cons:** much methods only for redirecting, need to call the needed methods manually (leads to bugs etc.)
- Dynamic composition with Optional:

``` java
        interface BlockComponent {
            void onRightClick(GlowBlock);
            Optional<? extends TileEntity> createTileEntity(GlowBlock);
            //...
        }

        class Openable extends AbstractBlockComponent {
            @Override void onRightClick(GlowBlock b) {...}
         }

        class Door extends AbstractBlockComponent { 
            @Override void onPlace(GlowBlock b) {...}
            @Override void onBreak(GlowBlock b) {...}
        }

        class WoodenDoor extends BlockType {
            WoodenDoor() {
                super(new Openable(), new Door());
             }
        }
```

**Pros:** Clean classes, easy to extend (just add an object to that constructor, no need to call some methods or such... all handled dynamically by BlockType.java)
 **Cons:** performance issues? (probably not); more complex BlockType class
- Dynamic composition II:
  
  ``` java
  interface BlockComponent {
      void createTileEntity(Setter<TileEntity> t);
      static interface Setter<T> {
          void set(T obj);
      }
  }
  // rest is same as above
  ```

**Pros/Cons:** Not really sure, personal I wouldn't do it, it just came into my mind

Personally I'd prefer option 2. I think active exchange would be great.

Tonodus:
This PR is based on SpaceManiacs [suggestions](https://github.com/GlowstoneMC/Glowstone/issues/536#issuecomment-63026359), but I think it's highly recommended to discuss the implemented component system, before any testing is done.

The diff is quite huge, but a great amount is only single word changes or moved code - f.e. DefaultBlockType#rightClickBlock comes from BlockType#rightClickBlock.

_(fixes #536)_

Glowkit: https://github.com/GlowstoneMC/Glowkit/pull/60


src/main/java/net/glowstone/block/itemtype/DefaultItemType.java
```diff
@@ -0,0 +1,74 @@
...
+    public final Material getMaterial() {
+        return Material.getMaterial(id);
+    }
+
+    protected ItemType[] getFeatures() {
+        return features;
+    }
+
+    protected <T> T throwDouble(T nullValue, T value) {
+        if (nullValue != nullValue)
```
SpaceManiac: wrong comparison?

turt2live:
!!! @Tonodus Doesn't compile :( 

(see travis)

Tonodus:
@SpaceManiac fixed
@turt2live yeah... this needs https://github.com/GlowstoneMC/Glowkit/pull/60

`GLOWKIT: 60`

turt2live:
I'm putting this on hold until various other block-related PRs are pulled, such as #249 (physics) and #284 (explosions).

turt2live:
@Tonodus This pull request is no longer mergeable. Please resolve the issue as soon as possible, thanks!

Tonodus:
@turt2live ps: the "on hold" isn't needed anymore.
@SpaceManiac any chance this get pulled once or can this be closed due to the [blocks branch](https://github.com/GlowstoneMC/Glowstone/tree/blocks)?

SpaceManiac:
@Tonodus I'd be interested in your feedback on the current state of things on the blocks branch, and how well you think the system shown there solves things.

turt2live:
@Tonodus Your pull request has been unmergeable for 3 days now. Please resolve the conflicts as soon as possible.

dequis:
I noticed the conflict, tried to merge without even knowing what this was about, and saw this

```
    public ItemFlintAndSteel() {
<<<<<<< HEAD
        super(
                new ItemTool() {
                    @Override
                    public boolean onToolRightClick(GlowPlayer player, ItemStack holding, GlowBlock target, BlockFace face, Vector clickedLoc) {
```

My reaction was "WTF" until i realized that this PR is another approach to what the blocks branch is doing.

I think mergeability is irrelevant since this is more of a proposal of how to design stuff.

Tonodus and SpaceManiac should talk about this stuff if they haven't done so yet. Personally I like what spacemaniac did with the blocks branch, but I also like the yaml stuff that tonodus added in #541, for example

turt2live:
Held: Pending analysis of block system within Glowstone.

Tonodus:
@SpaceManiac @turt2live 
It's been a while since I've responded here, but I didn't forget about this:
After I did much more investigation on SpaceManiac's work, I think this PR can be closed now.

His draft includes much more functionality than this PR does and seems to be more future-proof to me, especially when we're looking at Minecraft's new BlockState system or support of sponge one day, while I doesn't miss functionality added by this PR.

But the most important: If we only compare this PR and his `BlockBehavior`, our solutions are quite similar (cf. `ListBlockBehavior` and `DefaultBlockType`) - biggest difference is the way, to say that certain functionality is not provided. Throwing a `Signal` instead of returning `null` has some advantages and I don't see any disadvantages for now.

Finally, I tried some experiments suitable for my solution with SpaceManiacs branch. I didn't notice any missing functionality or other disadvantages due to his behavior-approach.

I'm sorry, I'm not able to provide improvements for the blocks-branch right now. If anyone thinks this PR should be re-opened, because my implementation has advantages I don't see right now, feel free to respond.

Issue #569, opened 2014-11-26
Natively support proxy data from LilyPad


turt2live:
Considering Glowstone now supports the proxy data from Bungee, it is only reasonably fair to support the proxy data from [LilyPad](http://www.lilypadmc.org/) as well.

_Note: LilyPad does not use plugin messages to transfer information between servers, unlike Bungee. Supporting this data is out of scope for Glowstone but is still explained at the end of this ticket._

The major difference between the Bungee data and the LilyPad data is that the Bungee data is terminated by `NUL` while LilyPad uses a JSON string. Both still use the same field on the handshake packet though, so that's something that's easy to work with.

For a visual representation, Bungee data would look something like this:
`localhost\0127.0.0.1\0c465b1543c294dbfa7e3e0869504b8d8\0[{"name":"textures","value":"...","signature":"..."}]`
...or in a more readable breakdown...
`Hostname \0 IP Address \0 UUID \0 [JSON array of properties]`

This varies from LilyPad quite a lot, as LilyPad data would look something like this (for the same player):
`{"h":"localhost","rIp":"127.0.0.1","rP":25565,"n":"turt2live","u":"c465b1543c294dbfa7e3e0869504b8d8","s":"...","p":[{"n":"textures","v":"...","s":"..."}]}`

As you can see, it's very different. LilyPad adds additional fields that Bungee doesn't have: Port number, player name, and a "secret" (`s`). The secret is used for internal operations to verify that the server being connected to is a valid Minecraft server on the proxy network.

Glowstone won't need to (and shouldn't) support the handling of the secret, but it should at least be made decently accessible for the LilyPad author(s) so that we can avoid getting in the way of their workflow.

Additionally, it should be noted that the "properties" format sent by LilyPad varies significantly. 

---

_The remainder of this ticket is a description on how LilyPad doesn't use plugin messages. Feel free to skip._

LilyPad operates off of 3 roles: Proxy, Server, and Connect. The "Connect" role is a central server that handles the various incoming connections and routes them as needed. The "Proxy" role is the server you connect to with your Minecraft client. The "Server" role is, in this case, Glowstone.

[...]

PR #570, opened 2014-11-26
Support LilyPad proxy information (adds #569)

turt2live:
Please refer to #569 for a technical description of the data format and why this is important.

---

This PR adds support for the proxy data sent by [LilyPad](http://www.lilypadmc.org/). This is done by attempting to parse the JSON data first. If parsing failed, then the data is likely Bungee or other (unsupported) data. If parsing passed on the other hand, then we rip it apart for our purposes. 

The Bungee parsing has only been changed to ensure that the two unsupported values are set to null. All other processing remains unchanged.

This PR also adds support for the security key LilyPad sends for it's internal use. This is not used by Glowstone at all but is included to make the lives of the LilyPad maintainers easier. By having this field exposed (to a degree) helps ensure that they don't have to keep up to date with our networking logic and rewrite the GlowProtocol used for handshakes. It will still require effort on their part to get access to this field, but at least it is much easier than having to override our networking.

It should also be noted that this PR technically allows the use of multiple proxies (ie: Bungee and LilyPad). Whether or not the independent assemblies permit that is another story, but the handling of the additional proxy data permits the use of both proxies.

### DIFF
--- a/src/main/java/net/glowstone/net/ProxyData.java
+++ b/src/main/java/net/glowstone/net/ProxyData.java
@@ -6,6 +6,7 @@ import net.glowstone.util.UuidUtils;
 import org.json.simple.JSONArray;
 import org.json.simple.JSONObject;
 import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
 
 import java.net.InetSocketAddress;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


turt2live:

turt2live:
Closing this PR - Project has moved on to a [new repo](https://github.com/GlowstoneMC/Glowstone). 

If this needs implementing in the newer repository I'll try and find time to implement it.

PR #577, opened 2014-11-28
Player Takes Void Damage

jaredStef:
When a player is deep in the void aka y axis is less than or equal to -64 the player takes 4 damage.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowEntity.java
@@ -296,27 +296,6 @@ public abstract class GlowEntity implements Entity {
     }
 
     /**
-     *
-     * @param material
-     * @return if the player is within 1 block of the material
-     */
-    public boolean isTouchingMaterial(Material material) {
-
-        /*todo: This isn't the most accurate way to do it. Once entity support is added I can make it more precise
-        use a bounding box
-        */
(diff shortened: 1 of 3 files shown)


turt2live:
`PRHOLD: Cacti hugging - contains additional required code`

jaredStef:
I won't submit a PR for suffocation yet either and hmm my build failed. Let me look at the checkStyle

dequis:
How does this depend on the [hugs PR](https://github.com/GlowstoneMC/Glowstone/pull/550)?

Also this seems like it would be nice to test timing of damage. From the mc wiki:

> Falling into the Void past Y=-64 causes the player's health to be depleted by about 4 (:heart::heart:) every half second, leading to the player's death in around 2.5 seconds. Armor does not protect you from dying in the Void, and death always occurs at Y=-199.10

turt2live:
It will not work without the boiler plate code asked in the hugs PR.

sent from mobile

dequis:
What boiler plate code? I see it adds canTakeDamage and isTouchingMaterial, neither of which are used or needed here

Although it does duplicate the `noDamageTicks = maxNoDamageTicks` line, but.. that's no reason to put it on hold.

turt2live:
Exactly: it doesn't make use of them. It should make use of them for the
sake of consistency throughout the code.

sent from mobile

dequis:
This doesn't need to be on hold anymore

SpaceManiac:
Merged manually in 80ee09c652579da034f576b3eda89864dd4a6a1a. Thanks!

Issue #586, opened 2014-12-06
Unintentionally broke MultiCraft (and some hosting provider) support


turt2live:
Now that Glowstone has a warning for not having a console (see 23a4b8237915d49c67f259e3e33a8f3f6cee19ca) Multicraft no longer functions.

This is due to Multicraft starting the jar in a process and not as a console application. Many other hosts also provide a similar starting system and therefore do not have console access. 

Ideally, commit 23a4b8237915d49c67f259e3e33a8f3f6cee19ca would have been tested more before being thrown into master to avoid the lack of research that was stated in #581. 

My personal suggestion is to implement a command line argument to prevent the warning. In addition, I would suggest to implement a check to determine if an X server (or similar) is available, if one is not then simply printing the warning before exiting would be best. Many hosts and panels, such as Multicraft, still provide a tail of the log and therefore a printed message would be suitable.

bendem:
I don't really think the default behavior should be to prevent the server from starting. Most of the server will be run from hosts taking care of that part or on a dedicated server (thus launched using the console anyway).

Using a flag to run glowstone anyway wouldn't make sense either. If the host doesn't provide a console nor an X server, they won't provide a way to add startup flags.

IMHO, preventing glowstone to run if no console is found is not the right call.

> My personal suggestion is to implement a command line argument to prevent the warning. In addition, I would suggest to implement a check to determine if an X server (or similar) is available, if one is not then simply printing the warning before exiting would be best.

If no X server has been found, I don't think there is a need to prevent the server from starting.

mastercoms:
I think it should warn the user, both using the GUI window and in the console log, but start the server anyway.

turt2live:
The point of this ticket is to make note of a failure regarding what was implemented due to #581, not to re-discuss the topic in #581. 

Although in my opinion #581 was resolved too quickly this is not the place to re-discuss the problem scope of another ticket.

SpaceManiac:
Okay, I've disabled it for now until we can work out a specific way to detect the console situation in a way that doesn't ruin anything for weird console environments.

mastercoms:
@turt2live So we should discuss in #581?

turt2live:
@mastercoms My personal opinion would be to discuss it on the forums instead. Reviving dead tickets is not a good idea.

mastercoms:
@turt2live Alright.

bendem:
> in a way that doesn't ruin anything for weird console environments

[...]

PR #587, opened 2014-12-07
Fixed console check on headless environments

johnnyapol:
Addresses #586 

What this does is re-enables and patches the console check to work on environments that are not headless (meaning they have a display, keyboard, or mouse) and if they support the Desktop AWT apis. 
This allows the application to run on headless servers but blocks people double clicking on windows the jar file.

This was tested on a Windows 8.1 (x64) machine running the latest Java 8 update (oracle JVM), with and without the -Djava.awt.headless=true flag.

EDIT:

This is my first PR to the Glowstone project, so forgive me If I have not followed protocol correctly.


turt2live:
This PR is small enough that it would be preferred to be discussed on IRC with SpaceManiac or on the ticket itself.

Thanks for your contribution however!

PR #572, opened 2014-11-26
Add more biome decorators

jimmikaelkael:
This is an extension to PR #474 (already pulled) to add more complete biome decorations so that it'll be ready when the biome grid will be functional.
##### Work done
- Under water block patches (sand, clay and gravel)
- Mushrooms decorator
- Trees decorator
- Improved ore populator (generates granite, diorite and andesite)
- Improved lake decorator
- Mossy stone boulder decorator
- Emerald ore decorator
- Monster eggs block populator
- Irregular bedrock layer (in order to diminish the valuable ores at the bottom)
- IceDecorator (ice patches and spikes for ice plains spikes)
- Big oak tree gen
- filled OverworldPopulator with all overworld biome variation values

Gallery link: http://imgur.com/a/YzRho

**Related Links:**
- Ticket: #76

Issue #76, opened 2014-02-11
World generation


SpaceManiac:
World generation is fairly bleh right now and does not feature biomes. It also does not use the updated generation API. The optimal solution to rectify this would be:
- Adapting world generation from Spout vanilla to either:
  - a) Include in core as the default world gen.
  - b) Distribute as official plugin, with an extremely minimal flatlands core gen.

Alternate options that were considered:
- (In plugin form ONLY - not core) Somehow linking with vanilla jar to get accurate terrain gen.
- Creating new world generation by regular contributor.
- Asking around for an existing Bukkit generator to include.

The generator should reasonably approximate the vanilla generator in terms of biome diversity and characteristics, and should not have obvious flaws.

zyuiop:
Can we use the NMS code and adapt it to GlowStone to implement this feature ?

TomCaserta:
100% no. NMS is copyrighted code and would cause major legal issues if any of the code was copied from decompiled source code. 

Nothing stopping people from viewing the code to get an idea of the implementation (ie to find out what each meta data bit does etc) however any sort of copying is out of the question.

EpicSharkNoob:
Can anyone tell me a world generation plugin that spawns only in-game biomes, and not some weird thing like "city" biome or "rainbow mountains" biome?  That would be AWESOME

jimmikaelkael:
This is way far to be a complete world gen, but I can help if needed. I already have implemented huge mushrooms in PR #374. After I will have completed this PR I think I can try to reproduce the different possible vanilla trees.

EpicSharkNoob:
Hi,

Thank you SO MUCH for the answers.  It really bugs me to spawn a
jungle world with ugly mountain-colored trees and grass.  :)

Also, in response to Michael J, I would LOVE some help!!  That would be
awesome.  Now I can get my Multiplayer Deathmatch server up!

Thank you both 1,000,000,000 times,

EpicSharkNoob

MeanMaggie:
Terrain Control plugin works well.  I use it on my Craftbukkit server and is compatible with Spigot, Cauldron and Forge.   Here is the link:  https://github.com/MCTCP/TerrainControl/wiki/Installation-instructions

mastercoms:
@MeanMaggie Except that uses nms, which is not in Glowstone

MeanMaggie:
[...]

PR #564, opened 2014-11-22
[WIP] Add some structures and their NBT storage

jimmikaelkael:
This PR will add some vanilla structures:
- Witch Hut
- Dungeons
- Desert Well
- Jungle Temple
- Desert Temple

**Related Links:**
- Ticket: #76

PR #607, opened 2015-01-11
Add overworld world generator

jimmikaelkael:
This merges PR #572 and #597 and add one commit on top of it:
https://github.com/GlowstoneMC/Glowstone/commit/888f7846be0727f1b74da6172811428fa0cc1742

You can see the results here:
http://imgur.com/a/huhmT 

---

**Related Links:**
- Ticket: #76


SpaceManiac:
I really love what you've done! The terrain generation looks just fantastic. I've reviewed the implementation a little bit and so far have a few questions:
- The entire ocean floor seems to be gravel (with stone only one layer below). I think this should be dirt instead, and for the first several layers instead of just one, to match vanilla more closely. Thoughts?
- DungeonPopulator's method for picking random tools is commented as if it expects index to be in the [0,3] range, but is called with a 4 index, leading to items with odd rarities:
  - Diamond rarity: STICK
  - Gold rarity: STRING
  - Iron rarity: FLINT_AND_STEEL
  - Stone rarity: DIAMOND_SWORD
  - Wood rarity: STONE_SWORD

jimmikaelkael:
@SpaceManiac OK I'll try to add several layers of dirt below ocean gravel. 
For dungeons, I think what you're talking about is in the actual implementation (which is not interfered with in this PR) which should be replaced once PR #564 pulled and so have the correct chest loots after that.

It still WIP anyway as I need #572 and #597 to be pulled first: it will let me some time to fix a few problems:
- Mesa doesn't have colored stripes (almost done)
- Bryce doesn't have spikes (almost done)
- Mesa forests doesn't have grass and coarse dirt (almost done)
- Switch to Simplex noise for flowers
- Needs several layers of dirt below ocean gravel as you mentionned
- In roofed forest there's not enough tall grass on ground (due to tree density)
- In jungles there's not enough ground bushes
- Caves and Ravine generator isn't done (not planned for this PR)
- Block lighting isn't working and once operational it should overall look better (Not in the scope of this PR)
- Biome temperature is not handled so water placed by the generator is never turned into ice (Not in the scope of this PR)

Another thing that could be cool to refine world gen would be to have AMIDST compatible with Glowstone so that we can check the world map in real time.

jimmikaelkael:
@SpaceManiac I checked it once again on a fresh vanilla 1.8 generated map in solo and it's stone directly under gravel in oceans.
Possibly you can expect to find some dirt patches here and there because it's populated in stone as an ore.

SpaceManiac:
Ah, it would seem you are correct on both counts. I was operating on outdated knowledge wrt the ocean floor - you can keep it as-is - and indeed your PR does not touch the DungeonPopulator (I had been reviewing the entire generator package).

jimmikaelkael:
I think I have handled most of the problems mentionned above, leaving 2 of them for another PR:
- Caves and Ravine generator isn't done
- Block lighting isn't working and once operational it should overall look better

gdude2002:
This looks pretty fantastic, IMO. Is it done? Can't wait to have this in master ._.

jimmikaelkael:
I still need to adjust a few things. I'll notify when it's ready which should happen in the coming days ;)

gdude2002:
<3

jimmikaelkael:
I think it's ready.

Tonodus:
I've just reset my world to test your PR.... you've definitely done awesome work! :+1:

bennasar99:
Merge would be awesome.

gdude2002:
Is there anything holding this back? @SpaceManiac?

jimmikaelkael:
The two other PRs I guess,  #572 and #597.

turt2live:
@gdude2002 Pending review.

PR #613, opened 2015-01-17
Add support for configurable world height

ghost:
This implements the max world height parameter in the server configuration. Regardless of the max world height worlds can still have blocks up to Y=256 (exclusive) to stay consistent to the Anvil format.

I also corrected the sea level as it should be at a constant 64 blocks.

### DIFF
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -1642,4 +1642,8 @@ public final class GlowServer implements Server {
     public boolean getAllowFlight() {
         return config.getBoolean(ServerConfig.Key.ALLOW_FLIGHT);
     }
+
+    public int getMaxBuildHeight() {
+        return config.getInt(ServerConfig.Key.MAX_BUILD_HEIGHT);
+    }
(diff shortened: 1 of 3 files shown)


turt2live:
Is there a reason why you don't modify the `DEPTH` constant of `GlowChunk`? That constant is used for entity spawning, which if left at it's current value would mean that entities will not function correctly in the future.

ghost:
I left the `DEPTH` constant as-is because despite the height limit, worlds are always saved with a height 256 blocks, and allowing this to be lower could lead to worlds getting cut off if the height limit is lowered. It also could break compatibility with vanilla if the limit were raised higher than 256 and blocks were saved above 256.

turt2live:
So by making it configurable we risk losing data as is then...

turt2live:
If this were to be added it would need to be clamped to the range 64 - 256, inclusive.

turt2live:
@coasterman10 Any update on the changes requested?

ghost:
Added another commit, the maximum height will be clamped between 64 and 256.

Issue #75, opened 2014-02-11
Finish updating packets


SpaceManiac:
Finish ~~transferring packets from net.glowstone.msg to the new system and~~ adding newly created packets. Only the `PLAY` protocol is missing packets currently. Packets associated with low-priority or unimplemented features are themselves low-priority. Documentation is at http://wiki.vg/Protocol.

Outbound:
- [ ] 0A Use bed
- [ ] 14 Entity nothing happens (?)
- [ ] 20 Entity properties (?)
- [ ] 25 Block break animation (shows crack effect)
- [x] 27 Explosion (involves multi block change)
- [x] 35 Update Block Entity (mob spawner)
- [x] 3B-3E Scoreboards

SpaceManiac:
Commit 3b7d232ad168bd26bc9ef98add3c97e4bb0e2975 finishes updating the `net.glowstone.msg` package to the new system (though many of them are still unused). There are still several unimplemented packets remaining.

PR #619, opened 2015-01-22
Add bed interaction / sleeping

Johni0702:
This PR adds interaction with beds, sleeping and bed spawn points.
### Interacting and sleeping

When the player right-clicks on the bed and the following conditions are met, the player enters the bed:
- Biome is neither HELL nor SKY aka. End (bed will explode in these cases)
- It is night time or thundering
- The bed must not be occupied
- Player must be within reach (smaller than normal)
- No hostile mob must be within certain distance

This PR calls `PlayerBedLeaveEvent` and `PlayerBedEnterEvent` where necessary.
If every player in a world is sleeping for at least 100 ticks, the night is skipped and weather is reset.
### Bed spawn

If a player leaves the bed by either clicking 'Leave Bed' or completing sleep, their spawn is set to the head of the bed.
When the player respawns and they have a bed spawn set, this looks for the bed. If it is still there (or was destroyed and replaced) and a valid spawn location can be found next to the bed (in [the same order](https://i.johni0702.de/fR.png) as vanilla does), the player will spawn at that location instead of the world spawn.
If their bed was missing or obstructed, their spawn is reset and they'll spawn at the world spawn.
### Additional notes

This does not include removing beds in any way.
When the client leaves their bed, they determine their own position. Only respawning after death must be handled by the server. However this PR also handles the first case as the new location must be available to plugins when the `PlayerBedLeaveEvent` is called.
I could not find a pattern in the blocks the client tolerates when leaving their bed, therefore I've created a helper method `BlockBed.isValidSpawn(Material)`. If there is a more simple way to determine these block, which I failed to find, it should be used instead.
The world I used to create above screenshot and to test distances, exit positions, etc. can be downloaded [here](https://i.johni0702.de/a.tar.bz2). It was created in Vanilla 1.8 and might be useful for testing.
### Relevant links

This adds the UseBed packet: #75 
Glowkit: GlowstoneMC/Glowkit#66

### DIFF
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -371,19 +371,19 @@ public final class GlowWorld implements World {
      */
     public void pulse() {
         List<GlowEntity> temp = new ArrayList<>(entities.getAll());
-        List<GlowEntity> players = new LinkedList<>();
+        List<GlowPlayer> players = new LinkedList<>();
 
         // pulse players last so they actually see that other entities have
         // moved. unfortunately pretty hacky. not a problem for players b/c
         // their position is modified by session ticking.
         for (GlowEntity entity : temp) {
             if (entity instanceof GlowPlayer) {
-                players.add(entity);
+                players.add((GlowPlayer) entity);
             } else {
                 entity.pulse();
             }
         }
-        for (GlowEntity entity : players) {
+        for (GlowPlayer entity : players) {
             entity.pulse();
         }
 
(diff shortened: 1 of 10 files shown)


turt2live:
> This does not include removing beds in any way.
> Why not?

Johni0702:
@turt2live I figured that this PR is already big enough and removing beds could be done in a separate PR without any conflicts. But if you wish, I'll add bed removal as well.

turt2live:
What would be involved in removing a bed? Should it not be one method override?

Johni0702:
Yes, it should be rather straight forward. Override the method, remove the other half of the bed, eject any player and make sure all events fire as expected, etc.
I'll probably implement it anyways. It's up to you whether you want it in this PR or in a separate one.

turt2live:
In this one is preferred

Johni0702:
I finally got to have a look at bed breaking and it turns out that it's more complicated than I thought.
The main reason being that it sometimes drops an item and sometime doesn't (especially when setting the head/foot programmatically via `setType(Material)` etc.).
I haven't had the time to look at it closer but I'd rather have that done another time than blocking this PR.

Updated the Glowkit PR as well as this PR to include the required changes discussed in the Glowkit PR.

PR #621, opened 2015-01-30
Fix GlowBlock#getDrops() and #breakNaturally()

### LINKED ISSUE #620: Wrong implementation of GlowBlock#getDrops() and #breakNaturally()
In Glowstone's current implementation the methods mentioned above act like a player breaking the block with his hand.

For all blocks needing a special tool to mine...
- this results in wrong results for GlowBlock#getDrops() [without parameter]
- this results in no drops if GlowBlock#breakNaturally() is called.
-  => this causes to no block drops if the block is destroyed by an explosion.

Broken by an explosion...
- stone should drop cobblestone
- iron/gold ore should drop their items
- diamond/restone ore should drop diamonds/redstone
- leaves should drop their saplings, apples, etc.
- etc.

So basically an explosion 'mines' blocks with the needed tool; the mentioned methods should act the same.

Tonodus:
This fixes #620 (see for detailed explanation)

### DIFF
--- a/src/main/java/net/glowstone/block/GlowBlock.java
+++ b/src/main/java/net/glowstone/block/GlowBlock.java
@@ -330,7 +330,8 @@ public final class GlowBlock implements Block {
         }
 
         Location location = getLocation();
-        for (ItemStack stack : getDrops()) {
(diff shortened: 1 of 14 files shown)


turt2live:
Held: Pending analysis of block system within Glowstone.

PR #623, opened 2015-02-01
Attributes

### LINKED ISSUE #609: Player max health is not updated
Setting the max health of a player through `setMaxHealth` does not update the player's view of their max health as it should, leading to odd scaling for players with modified max health.

It might be that the max health must be conveyed through entity attributes (play outbound 0x20), which Glowstone does not yet support.

Tonodus:
This PR adds an AttributeStorage class, handling a LivingEntity's attributes.
Currently only used for maxHealth, needs further investigation for modifiers. Fixes #609

### DIFF
--- a/src/main/java/net/glowstone/entity/AttributeManager.java
+++ b/src/main/java/net/glowstone/entity/AttributeManager.java
@@ -49,9 +49,6 @@ public class AttributeManager {
         return key.def;
     }
 
-    //////////////////////////////////
-    // Internals
-
     public void setProperty(String key, double value, List<Modifier> modifiers) {
         if (properties.containsKey(key)) {
             properties.get(key).value = value;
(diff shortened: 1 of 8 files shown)


src/main/java/net/glowstone/entity/GlowCreature.java
```diff
@@ -22,7 +22,7 @@
      * The type of monster.
      */
     private final EntityType type;
-   
+
```
turt2live: Don't modify files that you do not touch otherwise.

src/main/java/net/glowstone/entity/AttributeManager.java
```diff
@@ -0,0 +1,111 @@
...
+            return properties.get(key.toString()).value;
+        }
+
+        return key.def;
+    }
+
+    //////////////////////////////////
+    // Internals
+
+    public void setProperty(String key, double value, List<Modifier> modifiers) {
```
ZephireNZ, line 52: If these are internal methods, why are they public?

PR #624, opened 2015-02-02
Allow using Glowstones Io utils without requiring a world

PaulBGD:
Essentially, Glowstone's IO utils can be used to load a world without creating a server. The one problem that arises, is that the GlowBlock class assumes there is a server/world. This cancels sending update packets to players if the block is not in a server/world.


turt2live:
How does this apply to Glowstone's operation as it stands? There should be a referenced ticket for this describing the concern in detail alongside information on this PR that supports this change as being correct and ideal for example situations and beyond.

PaulBGD:
Alright, I'll do the ticket tomorrow.

PaulBGD:
Fixes #625

Issue #462, opened 2014-10-20
Tool durability


dequis:
http://minecraft.gamepedia.com/Item_durability
- [x] Internal tracking of durability points
- [ ] Reduce points for each "effective use":
  - [ ] Breaking blocks
    - [ ] Pickaxe, axe or shovel
    - [ ] Sword (two points)
  - [x] Attacking entities
    - [x] Sword
    - [x] Pickaxe, axe or shovel (two points)
  - [ ] Specific actions of the following tools (they work as bare hands for breaking/attacking)
    - [ ] Hoe
    - [ ] Carrot on a Stick
    - [ ] Flint and Steel
    - [ ] Fishing Rods
    - [ ] Shears
    - [ ] Bows
- [ ] Set base durability points of items:
  - [ ] Pickaxe, axe shovel, sword, hoe - by material:
    - [ ] Gold (33)
    - [ ] Wood (60)
    - [ ] Stone (132)
    - [ ] Iron (251)
    - [ ] Diamond (1562)
  - [ ] Tools with their own durability:
    - [ ] Carrot on a Stick (26)
    - [ ] Flint and Steel (65)
    - [ ] Fishing Rods (65)
    - [ ] Shears (239)
    - [ ] Bows (385)
- [ ] Unbreaking enchantment ("_for a given enchantment level, the chance that they will use durability is 1 in (1+level). The result is that they will last an extra level times their original durability, give or take a few_")

turt2live:
Attacking is implemented in #461 (for ref)

SpaceManiac:
The base ("maximum") durability for all the tools listed are currently available through `org.bukkit.Material`.

ZephireNZ:
I've got an implementation of the base durability in #445, which adds on a field to ItemType. There's a little bit of bulk, but I've tried to optimise as much as possible - might not be perfect.

PR #445, opened 2014-10-16
Dynamic Recipes

ZephireNZ:
This provides the implementation side to GlowstoneMC/Glowkit#34. It makes banners, with others yet to come, craftable without creating countless entries in the static file.

Recipes to implement:
- [x] Banners
  - [x] Layer adding
  - [x] Duplication
- [x] Fireworks - Requires #506 for implementation
  - [x] Stars creation/editing
  - [x] Firework creation
- [x] Book duplication
- [x] Tool/Armor repair
- [x] Armor dyeing - Requires #496 for implementation
- [ ] Maps
  - [ ] Map zooming
  - [x] Map duplication

PR #627, opened 2015-02-16
Tool durability

### LINKED ISSUE #622: Player data not saved on server shutdown
**Bug description:**
Currently when the server is stopped no player data is saved (f.e. inventory contents).

**How to reproduce:**
Give yourself an item into your inventory.
Type `/stop` without disconnection! Let the server kick you.
Restart server and join.
=> your previously given item is lost

**Cause:**
Bukkit's stop command calls Bukkit.shutdown() => [GlowServer#shutdown()](https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/GlowServer.java#L601) => [every player is kicked](https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/GlowServer.java#L614) => [session gets disconnected](https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/entity/GlowPlayer.java#L1271) => [the channel is closed without further saving](https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/net/GlowSession.java#L355)

ghost:
Hi,
In this PR, base durability for Pickaxe, Axe, Shovel, Sword, Hoe is set. Reduce points on breaking blocks and using hoe. ~~And fix #622 too.~~

CrabeMan.


gdude2002:
#622 doesn't have anything to do with this PR - it would have been better to submit that fix as a separate PR.

#445 also has some base durability stuff. #462 is the relevant ticket for this PR.

Tonodus:
IMO your current approach is not ideal. A block shouldn't check the tools it's mined with, using multiple instanceof and type checks to apply damage to that tool.

I'd prefer a different attempt: Add a method `onBreakBlock(Player, GlowBlock, ItemStack)` to ItemType and override it in ItemTool, which checks for creative player and calls a new method like `calculateBreakDamage(GlowBlock)`. That method returns 1 by default, but is overriden by ItemSword to return 2 and by ItemHoe to return 0.

BTW ItemAxe, ItemPickaxe and ItemSpade are pretty useless. You might consider to remove them.

---

> #622 doesn't have anything to do with this PR - it would have been better to submit that fix as a separate PR.

Seems like a one-line-change, time for @SpaceManiac?

turt2live:
@CrabeMan Only one issue can be resolved per pull request. Remove one of the implementations for this to be reviewed.

Thanks.

src/main/java/net/glowstone/block/ItemTable.java
```diff
@@ -226,6 +226,19 @@ private void registerBuiltins() {
         reg(Material.ACACIA_DOOR_ITEM, new ItemPlaceAs(Material.ACACIA_DOOR));
         reg(Material.DARK_OAK_DOOR_ITEM, new ItemPlaceAs(Material.DARK_OAK_DOOR));
         reg(Material.WRITTEN_BOOK, new ItemWrittenBook());
+        reg(Material.IRON_SWORD, new ItemSword(Material.IRON_SWORD));
```
turt2live: Where are the iron spades, axes, and pickaxes?

src/main/java/net/glowstone/block/itemtype/ItemTool.java
```diff
@@ -17,23 +19,27 @@ public ItemTool(int maxDurability) {
...
         if (onToolRightClick(player, holding, target, face, clickedLoc)) {
-            damageTool(player, holding);
+            damageTool(player, holding, 1);
         }
     }
 
-    private void damageTool(GlowPlayer player, ItemStack holding) {
-        if (player.getGameMode() == GameMode.CREATIVE) {
+    protected void damageTool(GlowPlayer player, ItemStack holding, int damage) {
+        if (player.getGameMode() == GameMode.CREATIVE || damage == 0) {
```
turt2live: Why `damage == 0`?

src/main/java/net/glowstone/block/itemtype/ItemType.java
```diff
@@ -128,7 +128,17 @@ public void rightClickBlock(GlowPlayer player, GlowBlock target, BlockFace face,
...
-
+    
+    /**
+     * Called when a player break a block.
+     * @param player The player using the tool
+     * @param target The block right clicked with the tool
+     * @param holding The ItemStack the player was holding
+     */
+    public void onBreakBlock(GlowPlayer player, GlowBlock target, ItemStack holding) {
+     // nothing by default
```
turt2live: Minor concern: Indentation

src/main/java/net/glowstone/block/itemtype/ItemTool.java
```diff
@@ -17,23 +19,27 @@ public ItemTool(int maxDurability) {
...
             return;
         }
 
-        holding.setDurability((short) (holding.getDurability() + 1));
-        if (holding.getDurability() == maxDurability + 1) {
+        holding.setDurability((short) (holding.getDurability() + damage));
+        if (holding.getDurability() == maxDurability + damage) {
             EventFactory.callEvent(new PlayerItemBreakEvent(player, holding));
-            holding.setAmount(0);
+            player.getInventory().remove(holding);
```
turt2live, line 52: Reason for this change?

src/main/java/net/glowstone/block/itemtype/ItemHoe.java
```diff
@@ -2,20 +2,26 @@
...
+        super(material.getMaxDurability());
+    }
+    
     @Override
     public void rightClickBlock(GlowPlayer player, GlowBlock target, BlockFace face, ItemStack holding, Vector clickedLoc) {
         if (target.getRelative(BlockFace.UP).getType() == Material.AIR) {
             if (target.getType() == Material.GRASS || (target.getType() == Material.DIRT && target.getData() == 0)) {
                 // grass or ordinary dirt: become soil
                 target.getWorld().playSound(target.getLocation().add(0.5D, 0.5D, 0.5D), Sound.STEP_GRAVEL, 1, 0.8F);
+                damageTool(player, holding, 1);
```
turt2live: If you declare a method to hide constants, use it.

ghost:
Hello, yes i forgot Iron Spade, Pickaxe and Axe, sorry.
I added `damage == 0` because if damage inflicted on tool is equal to 0, we don't need to update durability on it. I have change `holding.setAmount(0)` by `player.getInventory().remove(holding)` because `setAmount(0)` don't remove item.
And for ItemHoe, which one constants do you mean?

CrabeMan.

turt2live:
You override your new method to return `0` and use a hardcoded `1`.

ghost:
Hello, yes the override method in ItemHoe `calculateBreakDamage` return damage for breaking block. But in `rightClickBlock` it's damage for using the Hoe.

CrabeMan.

ghost:
I have implemented Shears. But the time for break a leave with a shears is really fast and `DiggingHandler.handle` don't call `DiggingMessage.FINISH_DIGGING`. I fix it temporary, but it's horrible and other cases like this one is possible. Maybe Netherrack with Diamond Pickaxe Efficiency V. If someone has an idea of how can i do that more smartly.

Thanks in advance.
CrabeMan

turt2live:
Your PR does not compile.

ghost:
Fixed.

hibo98:
Small hint: The gold tools are missing.

ghost:
@hibo98 Yes thanks, i will add them this night when i'm back.

CrabeMan.

src/main/java/net/glowstone/block/itemtype/ItemShears.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class ItemShears extends ItemTool {
+
+    public ItemShears(Material material) {
+        super(material.getMaxDurability());
+    }
+
+    @Override
+    public short calculateBreakDamage(GlowBlock target) {
+        return (short) (target.getType().equals(Material.LEAVES) ? 1 : 0);
```
turt2live: Don't shears loose durability on other blocks as well?

turt2live:
Which ticket does this cover?

gdude2002:
@turt2live I believe it relates to #462

ghost:
Currently, this PR cover:
- Reduce points
  - Breaking blocks
  - Attacking entities
  - Using Hoe
  - Using Flint and Steel (6c6b63e)
- Set base durability
  - Pickaxe, axe shovel, sword, hoe
  - Carrot on a Stick
  - Flint and Steel
  - Fishing Rods
  - Shears
  - Bows
- Unbreaking enchantment

CrabeMan.

PR #631, opened 2015-03-06
Adding cauldron

Tonodus:
This PR adds BlockCauldron:
- filling the cauldron with WATER_BUCKET
- filling water bottles with the cauldron
- bleaching LeatherArmor and Banner

Everything is tested with a inventory generated in vanilla minecraft (colored armor and banner with pattern).


src/main/java/net/glowstone/block/blocktype/BlockCauldron.java
```diff
@@ -0,0 +1,130 @@
...
+            block.setData((byte) (block.getData() - 1));
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    @Override
+    public Collection<ItemStack> getMinedDrops(GlowBlock block, ItemStack tool) {
+        return DROP;
```
Ribesg, line 123: As both your `Collection` and the `ItemStack` are mutable, it would be a good idea to just build a new `Collection` and a new `ItemStack` here. I'm not sure about this, but it's what I would do.

PR #632, opened 2015-03-15
Add Block Suffication

jaredStef:
Living Entities will now suffocate when within a solid block. There is a <code>isWithinSolidBlock</code> method that checks if the entity is in a solid block and if so they are damaged 1 every tick for suffocation.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowLivingEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowLivingEntity.java
@@ -348,9 +348,9 @@ public abstract class GlowLivingEntity extends GlowEntity implements LivingEntit
     }
 
     /**
-     * Checks if the entities head is stuck in a block.
+     * Checks if the entitie's head is stuck in a block.
      */
-    public boolean isWithinBlock() {
+    public boolean isWithinSolidBlock() {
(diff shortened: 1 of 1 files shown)


PaulBGD:
You uh.. added a method called isWithinBlock but call it with isWithinSolidBlock?

jaredStef:
Facepalm... I'll change it

turt2live:
Minor concerns (no testing done yet):
- Spelling in JavaDocs
- Indentation

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -139,6 +139,9 @@ public void pulse() {
             damage(4, EntityDamageEvent.DamageCause.VOID);
         }
 
+        if (isWithinSolidBlock())
+                damage(1, EntityDamageEvent.DamageCause.SUFFOCATION);
+
```
Ribesg: [You should always have curly braces around conditional statements' blocks.](http://stackoverflow.com/a/8020255)

gdude2002: [Always](https://www.imperialviolet.org/2014/02/22/applebug.html).

PR #642, opened 2015-06-11
START_DIGGING==FINISH_DIGGING for blocks that break on first hit. Fixes #629

### LINKED ISSUE #629: DiggingMessage.FINISH_DIGGING not call
Hello, `DiggingHandler.handle` don't call `DiggingMessage.FINISH_DIGGING` if the speed to break block is ~<0.05, based in [this table](http://minecraft.gamepedia.com/Breaking#Blocks_by_hardness)

CrabeMan.

Cmdr3nder:
This code calls the `FINISH_DIGGING` event for the block if the block is of a type that breaks on the first hit and thus never receives the `FINISH_DIGGING` event naturally.

Fixes #629


Tonodus:
This is a different approach than #541 to fix #384 (#629 is a duplicate of #384).

PR #647, opened 2015-10-22
Fixes #622

### LINKED ISSUE #622: Player data not saved on server shutdown

tsengia:
This is to fix issue #622. 
Simply has the player data being saved when the player is kicked.

The `saveData` method is called as a non-asynchronous save in order to save the player data before the server shuts down. With an asynchronous save during shutdown, the server will kill the save thread, thus not saving any player data.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -378,6 +378,17 @@ public final class GlowPlayer extends GlowHumanEntity implements Player {
         getServer().setPlayerOnline(this, false);
         super.remove();
     }
+    
+    public void remove(boolean asyncSave) {
+        knownChunks.clear();
+        chunkLock.clear();
+        saveData(asyncSave);
+        getInventory().removeViewer(this);
+        getInventory().getCraftingInventory().removeViewer(this);
+        permissions.clearPermissions();
+        getServer().setPlayerOnline(this, false);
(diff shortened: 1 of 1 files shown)


gdude2002:
I would suggest submitting this to [Glowstone++](https://github.com/GlowstonePlusPlus/GlowstonePlusPlus/) as well, since it's more active.

tsengia:
Alright, I didn't realize this project was so inactive. I added a pull request to Glowstone++.

